PR #11, opened 2015-09-29
Add support for release schema.

vitillo:


moztelemetry/filter_service.py
```diff
@@ -81,7 +82,7 @@ def get_daily_stats(self, from_date, to_date):
         for i in range(n_days + 1):
             submission_date = (from_date + relativedelta(days=i)).strftime("%Y%m%d")
             domain = self[submission_date[:-2]]
-            query = "select count(*) from {} where submissionDate='{}'".format(domain.name, submission_date)
+            query = "select count(*) from `{}` where submissionDate='{}'".format(domain.name, submission_date)
```
whd: Looks like the addition of [backticks](https://docs.aws.amazon.com/AmazonSimpleDB/latest/DeveloperGuide/QuotingRulesSelect.html) here is due to having dashes in some prefixes.

moztelemetry/filter_service.py
```diff
@@ -251,14 +252,7 @@ def publish(prefix):
...
                        dims["submissionDate"][:-2] in sdb:
-                        attributes = {"appName": dims.get("appName"),
-                                      "sourceName": dims.get("sourceName"),
-                                      "docType": dims.get("docType"),
-                                      "submissionDate": dims.get("submissionDate"),
-                                      "appVersion": dims.get("appVersion"),
-                                      "sourceVersion": dims.get("sourceVersion"),
-                                      "appUpdateChannel": dims.get("appUpdateChannel"),
-                                      "appBuildID": dims.get("appBuildId")}
+                        attributes = dims
```
whd: Maybe this variable is unnecessary now, since it is just `dims`. At any rate, I didn't find the definition of `schema.get_dimension_map` in this repository but I assume its results are of the same format as `attributes`.

moztelemetry/filter_service.py
```diff
@@ -28,10 +28,11 @@
 
 
 class SDB:
-    def __init__(self, prefix, months_retention=12, read_only=True):
+    def __init__(self, prefix, months_retention=12, read_only=True, s3_prefix=None):
```
whd: Side note: I think the plan is to provide the retention period for each data set in the sources.json metadata (but it isn't there now). `months_retention` could probably then be derived from that value.

whd:
This PR looks like it also removes support for the V2 schema, which per https://bugzilla.mozilla.org/show_bug.cgi?id=1179751 makes sense.


The PR looks like it does the right thing to me, but perhaps instead of taking an s3 prefix as argument, the program should take a "data set" name (e.g. `telemetry`, `telemetry-release`). The program could then look up the corresponding entry in `s3://net-mozaws-prod-us-west-2-pipeline-metadata/sources.json` to get the bucket prefix and schema location. In the current case this allows us to map `telemetry` to the `telemetry-2` prefix indirectly, but sources.json also has other useful information such as the metadata prefix and (eventually) the retention period for the data set.

moztelemetry/filter_service.py
vitillo: Correct.

moztelemetry/filter_service.py
```diff
@@ -289,89 +283,31 @@ def publish(prefix):
...
     if from_date and not to_date:
         to_date = datetime.now().strftime("%Y%m%d")
 
     if limit:
         limit = int(limit)
 
-    if schema_version != "v2" and schema_version != "v4":
-        raise ValueError("Unsupported schema version")
+    if dataset not in ["telemetry", "telemetry-release"]:
+        raise ValueError("Unsupported prefix")
```
whd: This should say `Unsupported dataset`.

whd:
Other than the error string, :+1:

vitillo:
I will deploy the new version once you have completed backfilling the submissions on S3.

PR #13, opened 2015-10-08
Parallelize SimpleDB index updating by date.

vitillo:


mreid-moz:
Looks good - one thing, I think you need to update `setup.py` to install the `joblib` library

PR #26, opened 2015-10-29
Skip bytes hackishly

mreid-moz:
Temporary workaround for problems caused by "bytes" fields (in the Telemetry case, it is the `submission` field containing gzipped bytes).


mreid-moz:
Note that this is basically the same workaround we used [here](https://github.com/mozilla/telemetry-server/commit/085e8450419649ebdf856404f4fb4263e006e4f5)

PR #28, opened 2015-10-30
Add strip_date function

poiru:
This makes it easier to deal with the various date formats in the pings.


vitillo:
This module doesn't seem the right place for small clean up functions. We might want to provide a larger set of clean up functionality in the future in a different namespace.

mreid-moz:
Maybe telemetry-tools in the "util" dir?

vitillo:
Yes, seems like a good fit.

PR #37, opened 2015-11-18
Allow Histograms.json URL to be specified with get_pings_properties

poiru:
This is needed because we occasionally need to use histograms that have
been deleted from the mozilla-central Histograms.json.

### DIFF
--- a/moztelemetry/histogram.py
+++ b/moztelemetry/histogram.py
@@ -17,6 +17,9 @@ import ujson as json
 from functools32 import lru_cache
 from expiringdict import ExpiringDict
 
+HISTOGRAMS_JSON_REVISION = "https://hg.mozilla.org/mozilla-central/rev/tip"
+HISTOGRAMS_JSON_PATH = "/toolkit/components/telemetry/Histograms.json"
+
 # Ugly hack to speed-up aggregation.
 exponential_buckets = histogram_tools.exponential_buckets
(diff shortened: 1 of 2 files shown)


poiru:
@vitillo I decided to go with `histograms_url` (instead of `histograms_rev`) so that you can 1) copy paste the URL into your browser and 2) use a custom Histograms.json.

(I also tested this on a Spark cluster so I hopefully I don't break everything again!)

vitillo:
You shouldn't change the current API as other projects depend on it, e.g. the [aggregator](https://github.com/mozilla/python_mozaggregator/blob/master/mozaggregator/db.py#L84) job. Any change you do should be backward compatible.

poiru:
@vitillo OK, how about this? `histograms_url` takes precedence, but `revision` is still supported.

vitillo:
I would like to avoid giving one precedence over the other, if both params are specified we should throw an exception as it doens't make much sense.

poiru:
@vitillo Done.

vitillo:
Looks good, have you tested it with the "Hello world notebook" as well?

poiru:
@vitillo Yep. I also verified that `EVENTLOOP_UI_LAG_EXP_MS` works when using the Aurora Histograms.json.

vitillo:
Nice, thanks!

vitillo:
I deployed the new version pypi.

PR #38, opened 2015-11-26
Add additional_histograms parameter to support arbitrary histograms

poiru:
Tested by running Hello world analysis and a separate analysis that uses the `additional_histograms` parameter.


vitillo:
Could you add some documentation for the get_pings_properties function and Histogram constructor?

poiru:
@vitillo Done (I documented the histogram parameters).

vitillo:
Deployed

PR #42, opened 2016-02-12
Allow custom keys in output of get_pings_properties

dzeber:
It is convient to be able to rename the property keys in the output of get_pings_properties rather than using the full original paths. 

The functionality for separating the property names in the results from the paths is already there, and this change exposes it via the API.


vitillo:
Looks good, thanks!

PR #44, opened 2016-02-12
Handle case of passing "*" as an arg to get_pings()

dzeber:
The telemetry schema uses `"*"` as a wildcard, to mean "allow all values for that field". When this is passed to `get_pings` it should be interpreted as "match any value" but it is treated as "match the string `'*'`".

This can be fixed by replacing `"*"` with `None` right before calling `SDB.query`.

### DIFF
--- a/moztelemetry/spark.py
+++ b/moztelemetry/spark.py
@@ -349,6 +349,8 @@ def _get_filenames_v2(**kwargs):
         tk = translate.get(k, None)
         if not tk:
             raise ValueError("Invalid query attribute name specified: {}".format(k))
+        if isinstance(v, basestring) and v == "*":
+            v = None
         query[tk] = v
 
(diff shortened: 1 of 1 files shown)


moztelemetry/spark.py
```diff
@@ -349,6 +349,8 @@ def _get_filenames_v2(**kwargs):
         tk = translate.get(k, None)
         if not tk:
             raise ValueError("Invalid query attribute name specified: {}".format(k))
+        if isinstance(v, basestring) and v == "*":
```
vitillo: Why are you checking for the type of v before comparing it?

dzeber:
I guess this is not necessary. `v` is expected to be either string, tuple of strings, or `None`. I removed the check.

vitillo:
Looks good, thanks.

PR #46, opened 2016-02-13
Make get_pings_properties return all subhistograms for a keyedHistogram with missing key

dzeber:
If one of the paths passed to `get_pings_properties` points to the name of a keyedHistogram without including an actual key, return a dict containing subhistograms for all available keys for that keyedHistogram.

As far as I can tell, the only thing this could potentially break is if something depends on catching the error thrown at https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/spark.py#L484, although I'm guessing this would be unlikely.

### DIFF
--- a/moztelemetry/spark.py
+++ b/moztelemetry/spark.py
@@ -117,6 +117,11 @@ def get_pings_properties(pings, paths, only_median=False, with_processes=False,
     Returns a RDD of a subset of properties of pings. Child histograms are
     automatically merged with the parent histogram.
 
+    If one of the paths points to a keyedHistogram name without supplying the
+    actual key, returns a dict of all available subhistograms for that property.
+
+    :param with_processes: should separate parent and child histograms be
+                           included as well?
     :param histograms_url: see histogram.Histogram constructor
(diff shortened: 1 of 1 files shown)


moztelemetry/spark.py
```diff
@@ -431,12 +436,30 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
...
-                result[k] = v.get_value(only_median) if v else None
+            if path[0] == "keyedHistograms" and len(path) == 2:
+                # Include histograms for all available keys.
+                # These are returned as a subdict mapped from the property_name.
+                kh_keys = cursor["keyedHistograms"][path[1]].keys()
+                if kh_keys:
+                    kh_histograms = {}
+                    for kh_key in kh_keys:
+                        props = _get_merged_histograms(cursor, kh_key,
+                                                    path + [kh_key],
```
vitillo: nit: fix indentation

dzeber:
Fixed indentation and merged in previous PR change.

vitillo:
Looks good, thanks.

PR #47, opened 2016-02-18
Fix for _get_pings_properties for subhistograms

dzeber:
Patch for PR #46.

Sorry - rookie mistake. I forgot to handle the case missing keyedHistogram name. This showed up in use on a larger dataset.


vitillo:
I just deployed it on a.t.m.o.

PR #52, opened 2016-03-03
Add a standard function for converting day number to a date.

mreid-moz:
Convert a number (or numeric string) containing a number of days
since Jan 1, 1970 to a date.

Incidentally, also fix a line-break problem.

### DIFF
--- a/moztelemetry/standards.py
+++ b/moztelemetry/standards.py
@@ -8,19 +8,38 @@
 """ This module implements some standard functionality based on Telemetry data.
 """
 
-from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


vitillo:
Why are we storing in the pings the number of days since the epoch instead of a date?

mreid-moz:
I have no idea.

vitillo:
Nobody wants really to work with this format, we should file a client bug to use an ISO date instead.

mreid-moz:
We'll still need to deal with it from Old firefoxes for some time, so this
code could still be useful.

vitillo:
Fair enough. r+

PR #53, opened 2016-03-03
Increment package version for push to PyPI

mreid-moz:


mreid-moz:
Pushed to [PyPI](https://pypi.python.org/pypi/python_moztelemetry).

PR #55, opened 2016-03-04
Fix retrieving keyed histograms missing keys

Uberi:
When keyed histograms have keys in the child that don't exist in the parent, `get_ping_properties` doesn't return those keys.

Keys in keyed histograms in the child payloads are only returned if the keyed histogram in the parent process has those keys as well.

Here's a sample with the new hello world notebook: https://gist.github.com/Uberi/7d607a6035116f2f8c09. The second half was taking really long, so I interrupted it after running all the code that can call or indirectly access `_get_ping_properties` and `_get_merged_histograms`.


vitillo:
Looks good, thanks!

PR #56, opened 2016-03-08
Fix threading issue with SimpleDB

Uberi:
Calling `get_pings` in a loop will eventually (around 100 for the default hadoop setup) make us run out of threads.

This is because the threadpool for SDB shuts down asynchronously, so the number of threadpools keeps growing.

To solve this, we can make thread pool shutdown synchronous (to make the class behave correctly), and also cache the SDB instances (to make things fast again). We can cache these instances because they don't have any real mutable state.

Here's a notebook that demonstrates a loop that uses only one thread pool at a time: https://gist.github.com/Uberi/9e2315e152b35595e917

### DIFF
--- a/moztelemetry/filter_service.py
+++ b/moztelemetry/filter_service.py
@@ -71,6 +71,17 @@ class SDB:
         for domain_name in domains:
             self._domains[domain_name[len(self.prefix) + 1:]] = self.sdb.get_domain(domain_name)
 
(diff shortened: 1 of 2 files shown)


vitillo:
It looks good to me, have you tested the `get_pings` API with your changes?

vitillo:
r+

PR #57, opened 2016-03-11
add else block to avoid UnboundLocalError

SamPenrose:
As the comment says. It's possible we want to assign a different value.


vitillo:
[Fixed](https://github.com/mozilla/python_moztelemetry/commit/0e1b13d3bc3781806b4e31fcb363a828b2f6a9be), thanks.

PR #60, opened 2016-03-23
Standards crc32

SamPenrose:
standards.in_sample(ping, divisor) tests whether a ping's clientId puts it in the crc32-based 1% sample, fixing a bug noticed by @mreid-moz (Python's crc32 routine is signed but our Lua library is apparently unsigned). I added 0.1% and 10% as subset/superset of 1%; LMK if you want a different approach. 0.1% is in use.

### DIFF
--- a/moztelemetry/standards.py
+++ b/moztelemetry/standards.py
@@ -7,7 +7,7 @@
 
 """ This module implements some standard functionality based on Telemetry data.
 """
-
+import binascii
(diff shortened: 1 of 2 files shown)


mreid-moz:
I was hoping for something as simple as `get_sample_id(client_id, range=100)` that we could be sure would give the same value we get from the `sampleId` field in the data.

PR #62, opened 2016-04-08
Remove Heka file chunking.

vitillo:
This is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1248845.
Chunking was introduced to reduce the memory pressure. Since then
some configuration changes have landed that deal with the issue
in a more general way so that chunking should be no longer required.


mreid-moz:
Looks good.

PR #63, opened 2016-04-21
Setup tests

maurodoglio:
This adds a basic setup to run tests using [pytest](http://pytest.org).
It also adds histogram_tools.py to this repository, so from now on we need to bump up the minor version of this library everytime that file is updated on mozilla-central.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,12 @@ password:example_pass
 ```
 - Fetch the latest code with `git pull`
 - Update PyPI with `python setup.py sdist upload`
+
+
+## Updating histogram_tools.py
+moztelemetry/histogram_tools.py is a mirror of its counterpart from
+[mozilla-central](https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/telemetry/histogram_tools.py).
+To update it to the latest version you can run
+```bash
+bin/update_histogram_tools
+```
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: python
+python:
+  - "2.7"
+cache: pip
+install:
+  - wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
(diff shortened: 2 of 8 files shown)


vitillo:
OK, let me know when this is ready to be reviewed/merged.

maurodoglio:
We have a green run!
Caching the pip packages will make the travis run go from [7'45"](https://travis-ci.org/mozilla/python_moztelemetry/builds/125574800) to  about [1'](https://travis-ci.org/maurodoglio/python_moztelemetry) !

vitillo:
Nice! It seems that currently `histogram_tools.py` is not fetched when the package is built. Could you add a hook for it?

maurodoglio:
@vitillo I added histogram_tools.py to the repo in the first commit. It should be there when you install the package.

vitillo:
Right, I was just wondering if we should setup a hook that fetches the current `histogram_tools.py` from mozilla-central when the package is being tested locally. Maybe we can address this in another PR.

maurodoglio:
The second commit adds a script to update histogram_tools.py from mozilla-central.

README.md
```diff
@@ -14,3 +14,10 @@ password:example_pass
 ```
 - Fetch the latest code with `git pull`
 - Update PyPI with `python setup.py sdist upload`
+
+
+## Updating histogram_tools.py
+moztelemetry/histogram_tools.py is a mirror of its counterpart from
+[mozilla-central](https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/telemetry/histogram_tools.py).
+To update it to the latest version you can run `update_histogram_tools`.
+Such command is available only if you install the moztelemetry package using `python setup.py install`.
```
vitillo: Does this work with pip's [editable install](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs) as well?

maurodoglio: Yeah it works with editable installs as well. I'll update the readme.

maurodoglio: As an alternative, I can wrap the update_histogram_tools in a python standalone script and put it in a bin directory at the root of the repository. Do you have a preference?

vitillo: Yeah, your that makes more sense to me as the script is something we want to use only during development.

vitillo:
Looks good, thanks!

PR #81, opened 2016-09-22
Bug 1304693 - Better handle datasets with many small files

maurodoglio:
For most of the datasets grouping the source files by cumulative size
works well. For datasets with many small files, there may be cases where
the size threshold is never/hardly reached and all the files end up being in
a number of groups < defaultParallelism. Let's ignore the grouping in
those case.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/81)


vitillo:
Looks good, thanks!

PR #82, opened 2016-09-27
Bug 1305087 - Add HLL feature

jezdez:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/82)

### DIFF
--- a/moztelemetry/hll.py
+++ b/moztelemetry/hll.py
@@ -1,0 +1,72 @@
+"""
+Original code from spark-hyperloglog
+
+import com.mozilla.spark.sql.hyperloglog.aggregates._
+import com.mozilla.spark.sql.hyperloglog.functions._
+
+val hllMerge = new HyperLogLogMerge
+sqlContext.udf.register("hll_merge", hllMerge)
+sqlContext.udf.register("hll_create", hllCreate _)
+sqlContext.udf.register("hll_cardinality", hllCardinality _)
+
+val frame = sc.parallelize(List("a", "b", "c", "c")).toDF("id")
+val count = frame
+  .select(expr("hll_create(id, 12) as hll"))
+  .groupBy()
+  .agg(expr("hll_cardinality(hll_merge(hll)) as count"))
+  .show()
+
+
+
+The interface to the spark-hyperloglog API
+
+from pyspark.sql.types import *
+from moztelemetry import hll
+
+hll.register(spark)
+
+schema = StructType([StructField('id', StringType(), True)])
+df = spark.createDataFrame([("a",), ("b",), ("c",), ("c",)], ['id'], schema)
+count = df
+    .select(hll.create('id', 12).alias('hll'))
+    .agg(hll.cardinality(hll.merge('hll')).alias('count'))
+    .show()
+
+"""
+from pyspark.sql import SparkSession
+from pyspark.sql.column import Column, _to_java_column, _to_seq
+from pyspark.sql.types import BinaryType, LongType
+
+
+def hyperloglog():
+    session = SparkSession.builder.getOrCreate()
+    return session.sparkContext._jvm.com.mozilla.spark.sql.hyperloglog
+
+
+def register(session):
+    session.catalog.registerFunction(
+        'hll_merge', merge, returnType=BinaryType()
+    )
+    session.catalog.registerFunction(
+        'hll_cardinality', cardinality, returnType=LongType()
+    )
+    session.catalog.registerFunction(
+        'hll_create', create, returnType=BinaryType(),
+    )
+
+
+def merge(col):
+    session = SparkSession.builder.getOrCreate()
+    _hll_merge = hyperloglog().aggregates.HyperLogLogMerge().apply
+    return Column(_hll_merge(_to_seq(session.sparkContext,
+                  [col], _to_java_column)))
+
+
+def cardinality(col):
+    hllCardinality = hyperloglog().functions.package.hllCardinality
+    return Column(hllCardinality(col))
+
+
+def create(element, bits):
+    hllCreate = hyperloglog().functions.package.hllCreate
+    return Column(hllCreate(element, bits))


jezdez:
@maurodoglio @vitillo I have trouble understanding the Java tracebacks, and even knowing if they are red herrings or not. Can someone try this code locally?

Run pyspark like so: `pyspark --packages vitillo:spark-hyperloglog:1.1.1`

Then import `moztelemetry.hll` and register the spark session with it: `moztelemetry.hll.register(spark)`, where `spark` is a Spark 2.0 `SparkSession`.

moztelemetry/hll.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def hyperloglog():
+    session = SparkSession.builder.getOrCreate()
+    return session.sparkContext._jvm.com.mozilla.spark.sql.hyperloglog
+
+
+def register(session):
+    session.catalog.registerFunction(
+        'hll_merge', merge, returnType=BinaryType()
```
jezdez: Does `BinaryType` make sense her? It's what the `aggregates.HyperLogLogMerge` has as its [data type here](https://github.com/vitillo/spark-hyperloglog/blob/7a30b2efd7f4fa8c0e0f6e595ec600c37bb4fb19/src/main/scala/com/mozilla/spark/sql/hyperloglog/aggregates/HyperLogLogMerge.scala#L28)

moztelemetry/hll.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    session.catalog.registerFunction(
+        'hll_cardinality', cardinality, returnType=LongType()
+    )
+    session.catalog.registerFunction(
+        'hll_create', create, returnType=BinaryType(),
```
jezdez: Same for the create function, `BinaryType` is probably the right one, as it [returns that](https://github.com/vitillo/spark-hyperloglog/blob/7a30b2efd7f4fa8c0e0f6e595ec600c37bb4fb19/src/main/scala/com/mozilla/spark/sql/hyperloglog/functions/package.scala#L25-L28).

moztelemetry/hll.py
jezdez: Pyspark [maps byte arrays to that type](https://github.com/apache/spark/blob/7f16affa262b059580ed2775a7b05a767aa72315/python/pyspark/sql/types.py#L145-L149)

moztelemetry/hll.py
```diff
@@ -0,0 +1,72 @@
...
+val frame = sc.parallelize(List("a", "b", "c", "c")).toDF("id")
+val count = frame
+  .select(expr("hll_create(id, 12) as hll"))
+  .groupBy()
+  .agg(expr("hll_cardinality(hll_merge(hll)) as count"))
+  .show()
+
+
+
+The interface to the spark-hyperloglog API
```
jezdez: Here's a good example

moztelemetry/hll.py
vitillo: Yes it does.

moztelemetry/hll.py
vitillo: Correct.

moztelemetry/hll.py
```diff
@@ -0,0 +1,72 @@
...
+def merge(col):
+    session = SparkSession.builder.getOrCreate()
+    _hll_merge = hyperloglog().aggregates.HyperLogLogMerge().apply
+    return Column(_hll_merge(_to_seq(session.sparkContext,
+                  [col], _to_java_column)))
+
+
+def cardinality(col):
+    hllCardinality = hyperloglog().functions.package.hllCardinality
+    return Column(hllCardinality(col))
```
vitillo: These wrappers don't look right. As you can see `hllCardinality` takes as input an 'Array[Byte]' and returns a `Long`, while the way you wrapped it instead requires `hllCardinality` to take a `Column` as input and return a `Column` as output, which is pretty much what the built-in SparkSQL functions do, like e.g. [log](https://github.com/apache/spark/blob/de333d121da4cb80d45819cbcf8b4246e48ec4d0/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L1640). In other words we can't use the same approach used by SparkSQL builtins but we have to come up with something else. I don't know what that something else is though as I haven't ported Scala UDFs to Python before, so there is some investigation required to understand how that can be accomplished.

moztelemetry/hll.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    session.catalog.registerFunction(
+        'hll_cardinality', cardinality, returnType=LongType()
+    )
+    session.catalog.registerFunction(
+        'hll_create', create, returnType=BinaryType(),
+    )
+
+
+def merge(col):
```
vitillo: This wrapper looks ok.

moztelemetry/hll.py
vitillo: It's possible that for this code to work the `hllCardinality` function has to be wrapped (from the Scala side) with [`org.apache.spark.sql.functions.udf`](https://jaceklaskowski.gitbooks.io/mastering-apache-spark/content/spark-sql-udfs.html), and the wrapped function is the one you should use from Python. You might have to build a custom JAR of spark-hyperloglog to test this with e.g. `pyspark --jars foobar.jar`.

jezdez: Two things, I'm not following what you mean with the first paragraph about columns, I was mostly following this on top of the code sample shared by Saptarshi since I didn't know what input type is needed for this function to work. A byte array may be handled on the way back by the BinaryType, but on the way there, I assumed that'd simply be the return value of either `create` or `merge`. I'm not sure.. :(

I'm not sure if using wrapping the Scala function is the right approach here, since that's what registers a function on the side of Scala for use in SQL queries etc after all.

vitillo: I don't think Saptarshi's code did ever work so there is that.

maurodoglio:
I'm closing this for now.

PR #86, opened 2016-10-17
Bug 1302264 - Use metadata_prefix in the Dataset API

whd:
Similar to https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/spark.py#L277

This value of `metadata_prefix` is usually (but not always) the same as `prefix`, which is why it has worked so far. The case where it's not is when you have two datasets with the same metadata prefix, as occasioned by forthcoming dataset migrations. 

We should produce better documentation and tooling around `sources.json` in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/86)

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -207,6 +207,6 @@ class Dataset:
         except KeyError:
             raise Exception('Unknown source {}'.format(source_name))
 
-        schema = store.get_key('{}/schema.json'.format(source['prefix'])).read()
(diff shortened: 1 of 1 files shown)


autolander:
Autolander could not find a bug number in your pull request title. All pull requests should be in the format of: Bug [number] - [description].

maurodoglio:
Thanks @whd!

PR #88, opened 2016-11-08
Bug 1302264 - Add support for new-style gzip-compressed objects

whd:
Requires https://github.com/mozilla/telemetry-tools/pull/6.

I've made it the responsibility of the code that is returning the s3 byte stream to apply the streaming gzip wrapper, because that code generally has the easiest access to the object's content-encoding.

The implementation of `parse_heka_message` is mocked in the tests in this repo, but the gzip parsing code is tested in telemetry-tools.

The only piece without tests is the conditional movement of the payload from `Payload` to `Fields[submission]` in the new infra, which I have added a special case for and tested manually.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/88)


vitillo:
@maurodoglio r?

maurodoglio:
Hey @whd it looks like the tests are failing in CI, could you please fix it?

moztelemetry/store.py
```diff
@@ -36,7 +38,12 @@ def get_key(self, key):
         try:
             # get_key must return a file-like object because that's what's
             # required by parse_heka_message
-            return bucket.Object(key).get()['Body']
+            s3object = bucket.Object(key).get()
+            if s3object['ResponseMetadata']['HTTPHeaders'].get(
```
maurodoglio: According to the [boto3 docs](https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Object.get) there should be a `ContentEncoding` key in the returned dictionary

maurodoglio: It would be great to have a test for this condition.

whd:
@maurodoglio the travis build fails because it requires https://github.com/mozilla/telemetry-tools/pull/6 to be merged first. I'll have @mreid-moz review that because he has some knowledge of heka.

mreid-moz:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[moztelemetry/heka_message_parser.py, line 42 at r1](https://reviewable.io:443/reviews/mozilla/python_moztelemetry/88#-KW54_N39VnnQJsCzmYz-r1-42:-KWZK1gkpZCYg61AxgUq:b-oohav0) ([raw file](https://github.com/mozilla/python_moztelemetry/blob/5309c19954a96557099cc640f04c3fff412fb92c/moztelemetry/heka_message_parser.py#L42)):*
<details><summary><i>Previously, maurodoglio (Mauro Doglio) wrote…</i></summary>
> Someone  with some knowledge of heka should review these 4 lines. Maybe @mreid-moz ?
</details>
This looks fine to me. It could slightly change the semantics if there's a field in the `submission`/`payload` with the same name as one of the heka fields. Previously: the payload fields would be overwritten by any heka field with the same name. Now: which field wins out will be based on the order in which the fields are encountered.

In the telemetry use case, I don't think this is a concern, since we should never encounter a message where a field of the same name exists in both the `submission` and a heka message field.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/python_moztelemetry/88)*

moztelemetry/store.py
whd: I've updated the code to use the top-level key for content-encoding instead of the low-level one.

Like with telemetry-batch-view, the testing harness here creates an implementation of the `S3Store` (`InMemoryStore`) for testing that makes adding a test that actually uses the code above difficult. I can add a test like `test_get_gzip_key` that adds the streaming wrapper within the testing logic and confirms the data written to disk is decompressed, but that's essentially a subset of what the test in telemetry-tools does and to me feels like confirming that python can gzip and gunzip a file. It would also potentially involve changing the `S3Store` API to support specifying a `content-encoding` when uploading an object, which is something I didn't want to do. An option to avoid changing the API would be to use e.g. file extension to determine content-encoding for the `InMemoryStore`, but that reverts to the same  gzip-gunzip cycle.

whd:
I've come across an issue while testing this PR a bit more that requires investigation / discussion, so this PR should not be merged until it has been resolved. It's problematic enough that I will file a separate bug about it tomorrow, but the gist of it is that the client, the new infra, or both occasionally produce some large numbers that don't fit into doubles, and ujson doesn't like that very much. It is masked in the current infra by https://github.com/mpx/lua-cjson/issues/37.

whd:
I believe I've resolved the above issues, as follows:

Fall back to the standard python json parser for cases where ujson fails.

Per https://bugzilla.mozilla.org/show_bug.cgi?id=1326107 and @mreid-moz's above comment, perform the payload/submission merge in known order due to the residual NULLs.

@mreid-moz can you re-r? This still requires https://github.com/mozilla/telemetry-tools/pull/6.

whd:
I've imported/rebased the changes I had from mozilla/telemetry-tools/pull/6 here. I also added a UTF8 check per feedback in mozilla/telemetry-batch-view/pull/159 and added a helper json parsing method
since `json.loads` was being called (and failing) in multiple places.

I put the streaming gzip wrapper in a util/ subdirectory since it's not strictly related to heka but it can be moved elsewhere.

@maurodoglio / @mreid-moz can you re-r?

mreid-moz:
:lgtm:

---

Reviewed 9 of 9 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/python_moztelemetry/88#-:-KabjF1yyFV7RhWpgOQN:bnfp4nl)*

mreid-moz:
Looks good to me too.

PR #93, opened 2016-11-29
Bug 1321091 - Use '4' as the default sourceVersion

fbertsch:
I ran into quite a few problems switching the aggregates job to the Dataset API, because it took me awhile to figure out that get_pings is setting the sourceVersion to 4 as a default. I believe we should do that here as well.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/93)


maurodoglio:
We discussed this offline, we prefer to keep Dataset as generic as possible.

PR #94, opened 2016-12-13
silence perpetual complaint

SamPenrose:
Suggest we move to accepting the tragic loss of whitelist.json

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/94)


autolander:
Autolander could not find a bug number in your pull request title. All pull requests should be in the format of: Bug [number] - [description].

chutten:
The original purpose of the println was to put a line in someone's normal build log that could help explain to them why things were acting weird. I think it can still serve that purpose.

If there were some way to detect "are we running during a build or somewhere else?" then we could make the print conditional.

mreid-moz:
The `histogram_tools.py` file should be an exact copy of [the one in the Firefox source](https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/telemetry/histogram_tools.py).

It appears that it's not quite up-to-date at the moment, but I don't think we want it to purposely diverge. 

Can we just add a dummy `histogram-whitelists.json` file with contents of `{}` here?

SamPenrose:
I didn't understand the larger context. Happy to either add the dummy file or close the PR.

mreid-moz:
Can you try adding the file and see if the warning goes away (and if other things blow up)?

mreid-moz:
Tests don't pass with this change :(
Looks like the dummy whitelists file needs to be a little less dumb - per [the test log](https://travis-ci.org/mozilla/python_moztelemetry/builds/186193092) it should have a top-level "n_buckets" key with an empty array as value (and maybe placeholders for the other top-level keys [here](https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/telemetry/histogram-whitelists.json)).

SamPenrose:
Sigh. My apologies for opening this can of worms. I was able to get the test_histogram.py to not error out by putting this in the JSON file:
{"n_buckets": ["GC_REASON_2"]}
but python setup.py test hangs, even after I remove every other test file, and I don't know why. I am happy to close this PR out as not worth the trouble or hand it off to someone else ... LMK.

mreid-moz:
Agreed that this seems like more trouble than it's worth.

PR #108, opened 2017-01-30
Bug 1322768 - Use Spark to list file summaries

maurodoglio:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/108)

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -137,7 +137,7 @@ class Dataset:
 
             return self._scan(dimensions[1:], matched, clauses, executor)
 
-    def summaries(self, limit=None):
+    def summaries(self, sc, limit=None):
         """Summary of the files contained in the current dataset
 
         Every item in the summary is a dict containing a key name and the corresponding size of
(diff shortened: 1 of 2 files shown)


moztelemetry/dataset.py
```diff
@@ -156,10 +156,8 @@ def summaries(self, limit=None):
 
         with futures.ProcessPoolExecutor(MAX_CONCURRENCY) as executor:
             scanned = self._scan(schema, [self.prefix], clauses, executor)
```
vitillo: How much of summaries is spent in scanning the directories vs listing the keys usually?

maurodoglio: I don't remember the numbers, but in most of the cases the long pole is listing the keys. It really depends on the filter though, the extreme cases (i.e. either no filters or only one filter on buildid) behave very differently.

maurodoglio: Btw before we merge this I want to run a benchmark to verify some numbers.

maurodoglio: I timed the execution of the first Dataset query in the hello world notebook and I can confirm the long pole is listing the files; 80% of the time is spent on it.
Unsurprisingly the spark implementation is slightly slower in some cases and faster in other cases. Basically when there are a lots of directories to list on, the spark implementation is faster on a multi node cluster. For example with the following query on a 10 nodes cluster, the execution time went down from 30s to 20s:
```python
pings = Dataset.from_source("telemetry") \
    .where(docType='main') \
    .where(submissionDate=lambda x: "20161201" < x < "20170201") \
    .where(appUpdateChannel="nightly") \
    .records(sc, sample=0.1)
```
On the other hand, some queries like
```python
pings = Dataset.from_source("telemetry") \
    .where(docType='main') \
    .where(submissionDate="20161101") \
    .where(appUpdateChannel="nightly") \
    .records(sc, sample=0.1)
``` 
take about 3 seconds more. Overall, I think having a progress bar is worth a small speed penalty you get in some cases.

vitillo: r+

PR #109, opened 2017-01-31
Bug 1335219 - Fix some missing fields

mreid-moz:
It appears the `payload.update(...)` call was causing some fields
to be dropped.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/109)


mreid-moz:
Don't merge yet, I'd like to add a test case for this.

mreid-moz:
I'd like to get a couple of sets of eyes on this... @whd and @maurodoglio would you mind taking a look?

Here's what I think was happening. The `payload` variable would start off empty.
It would be replaced by parsing the `submission` protobuf field (or msg.payload, both appear to work the same).
Then split-out JSON fields would be added to the `result` variable as they were encountered, some of which were `payload.whatever` fields.
The call at the end to `payload.update(result)` would replace the _dictionary's_ `payload` key with the one from `result`, which did not contain the original stuff contained in `submission`.

To make that clearer, here is an example:
```
>>> a = {"foo1": { "bar": 1 }, "foo2": { "bar": 2 }}
>>> b = {"foo1": { "baz": 1 }, "foo2": { "bar": 3 }}
>>> a.update(b)
>>> a
{'foo1': {'baz': 1}, 'foo2': {'bar': 3}}
```

Whereas what we really want `a` to contain is
```
{'foo1': {'bar': 1, 'baz': 1}, 'foo2': {'bar': 3}}
```

whd:
As there doesn't seem to be a good way of doing this without taking two passes, this approach makes sense.

The test only checks data decoded the hindsight way (which includes the NULLs), but the data that's currently being read is in the old format. I think the fact that it works with the newer data should be sufficient however.

mreid-moz:
I added a test that covers the "old-style" messages that have the leftover data in the protobuf `Payload` field as well (these were also affected, and should be fixed)

whd:
👍

moztelemetry/heka/message_parser.py
```diff
@@ -25,19 +25,34 @@ def parse_heka_message(message):
...
+            # split out.
+            if field.name == 'submission':
+                payload = _parse_json(field.value_bytes[0].decode('utf-8'))
+                break
+
+    if payload is None:
         payload = {}
 
+    payload["meta"] = {
+        # TODO: uuid, logger, severity, env_version, pid
```
maurodoglio: Let's file a bug to keep track of what needs to be done here

moztelemetry/heka/message_parser.py
```diff
@@ -61,14 +71,10 @@ def _parse_heka_record(record):
...
-            result["meta"][name[0]] = value[0] if len(value) else ""
+            payload["meta"][name[0]] = value[0] if len(value) else ""
         else:
-            _add_field(result, name, value)
-
-    # merge results back into the payload/submission
-    # payload may contain NULLed out structures in the new infra and must
-    # therefore be the receiver of the update
-    payload.update(result)
+            print "Adding field {}".format(field.name)
```
maurodoglio: Please remove this print

tests/heka/test_message_parser.py
```diff
@@ -59,3 +59,22 @@ def test_unpack_strict(data_dir):
...
+    with open(filename, "rb") as o:
+        for r in message_parser.parse_heka_message(streaming_gzip_wrapper(o)):
+            assert set(r.keys()) == top_keys
+            assert set(r["payload"].keys()) == payload_keys
+
+    filename = "{}/test_telemetry_snappy.heka".format(data_dir)
+    with open(filename, "rb") as o:
+        for r in message_parser.parse_heka_message(o):
+            assert set(r.keys()) == top_keys
+            assert set(r["payload"].keys()) == payload_keys
```
maurodoglio: Missing new line at the end of file

moztelemetry/heka/message_parser.py
mreid-moz: woops!

mreid-moz:
Addressed review feedback

PR #111, opened 2017-02-13
Bug 1255738 - Add Dataset.select for properties extraction

maurodoglio:


moztelemetry/dataset.py
```diff
@@ -70,6 +71,9 @@ class Dataset:
         schema = ['submissionDate', 'docType', 'platform']
 
         records = Dataset(bucket, schema) \\
+            .select('clientId',
+                    os_name='"environment.system".os.name',
```
vitillo: I am not sure why this is needed since we "stitch" the message back together in [here](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/heka/message_parser.py#L27). Ideally this should work without exposing the ugly raw structure of the message as our Python users are not aware of it.

maurodoglio: It wasn't needed indeed

tests/test_dataset.py
```diff
@@ -229,6 +289,23 @@ def test_records_limit_and_sample(spark_context):
...
+@pytest.mark.slow
+def test_records_selection(spark_context):
+    bucket_name = 'test-bucket'
+    store = InMemoryStore(bucket_name)
+    key = 'dir1/subdir1/key1'
+    value = '{"a": {"b": { "c": "value"}}}'
+    store.store[key] = value
+    dataset = Dataset(bucket_name, ['dim1', 'dim2'], store=store).select(field='a.b.c')
+    records = dataset.records(spark_context, decode=decode)
+    assert records.collect() == [{'field': 'value'}]
```
sunahsuh: Can we add another few lines here to test chaining another select to `dataset` and calling `records()` again?

moztelemetry/dataset.py
```diff
@@ -70,6 +71,9 @@ class Dataset:
         schema = ['submissionDate', 'docType', 'platform']
 
         records = Dataset(bucket, schema) \\
+            .select('clientId',
+                    os_name='environment.system.os.name',
```
vitillo: Might be worthwhile adding some more examples showcasing things that we couldn't do before, like accessing the content of an array or something.

maurodoglio: Good idea.

tests/test_dataset.py
maurodoglio: done!

moztelemetry/dataset.py
maurodoglio: I added this line to the example `'payload.threadHangStats[].hangs[].stack[0:2]` which shows how to select a slice of an array nested under 2 other arrays. I don't think you can do that with `get_pings_properties` 😸

maurodoglio:
This is ready for another look

.travis.yml
```diff
@@ -24,7 +24,7 @@ install:
   - conda env create -f environment.yml
   - source activate test-environment
   - pip install . --process-dependency-links
-  - pip install coverage
+  - pip install 'pytest>=3' coverage
```
wlach: Why are you setting the dependency here and not in setup.py?

wlach:
Also, the unit tests seem to be failing now. The most likely culprit is the new 60s timeout for unit tests, but I am not 100% sure.

wlach:
Since @maurodoglio was away, I fixed things up myself in #138.

PR #114, opened 2017-02-24
Add Zeppelin plotting helper

vitillo:
Zeppelin doesn't currently support Matplotlib out of the box so this is the only way to display a plot.


moztelemetry/zeppelin.py
```diff
@@ -0,0 +1,25 @@
...
+        from moztelemetry.zeppelin import show
+
+        fig = plt.figure()
+        plt.plot([1, 2, 3])
+        show(fig)
+    """
+    img = StringIO()
+    fig.savefig(img, format='svg')
+    img.seek(0)
+    print("%html <div style='width:600px'>" + img.buf + "</div>")
```
maurodoglio: It may be good to add a `width` parameter to let the user customize the image size (default to 600px sounds good to me).

vitillo:
@maurodoglio RFAL

PR #116, opened 2017-03-15
Bug 1347283 - Catch submission deserialization exceptions

maurodoglio:
This is to take care of errors like https://gist.github.com/maurodoglio/42bb80125ffc1003d5243cdd463936a3

### DIFF
--- a/moztelemetry/heka/message_parser.py
+++ b/moztelemetry/heka/message_parser.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import gzip
+import logging
 import ssl
 import struct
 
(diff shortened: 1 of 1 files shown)


moztelemetry/heka/message_parser.py
```diff
@@ -35,7 +38,10 @@ def _parse_heka_record(record):
             # because it contains null values for JSON fields that have been
             # split out.
             if field.name == 'submission':
-                payload = _parse_json(field.value_bytes[0].decode('utf-8'))
+                try:
+                    payload = _parse_json(field.value_bytes[0].decode('utf-8'))
```
mreid-moz: Should we also wrap the above call where we parse `message.payload`?

whd: We probably don't need to do this, since `message.payload` is only ever populated by something that is guaranteed to be (escaped) UTF8.

whd: It appears simply calling `.decode('utf-8', 'ignore')` may be the better solution. The test case to reproduce in the bug works and this would map more closely to what we're doing in scala.

whd:
Closing in favor of #117.

PR #118, opened 2017-03-16
 Sort pings returned from HBase in reverse chronological order

vitillo:


mreid-moz:
Can you make the order user-definable? My last use of this API actually benefited from the pings being in oldest-to-newest order.

vitillo:
> Can you make the order user-definable? My last use of this API actually benefited from the pings being in oldest-to-newest order.

@mreid-moz done, please review the changes.

maurodoglio:
It would be great if you could get rid of the merge commit.

PR #119, opened 2017-03-16
Test against Spark 2 on travis

vitillo:


wlach:
Looking into why this isn't working.

wlach:
I think I found the root cause and suggested a fix in #120, closing this.

PR #127, opened 2017-03-27
Bug 1349608 - Improve Dataset handling of small pings

maurodoglio:
The idea here is to use a greedy algorithm to balance the RDD partition
in terms of number of files (first) and total size of each partition.

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -19,25 +19,22 @@ from .store import S3Store
 MAX_CONCURRENCY = int(cpu_count() * 1.5)
 
 
-def _group_by_size(obj_list):
-    """Partitions a list of objects by cumulative size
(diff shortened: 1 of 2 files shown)


maurodoglio:
I ran a few comparison tests on 2 different queries:
- 10% of every ping with docType = OTHER
    - Time before change: 16min 40s
    - Time after change: 10min 43s (**-34.7%**)
- 10% of main pings for a given submissionDate
    - Time before change: 35min 20s
    - Time after change: 22min 47s (**-35.5%**)
As a bonus point the logic in the grouping function is now simpler (and more readable)

maurodoglio:
I forgot to mention, to make the comparison more correct, I bypassed the sampling provided by .records and instead I used the same set of pre-sampled summaries for both. I can provide the sampled data as json if anybody is interested.

PR #128, opened 2017-03-28
Bug 1318681 - Make Dataset sampling repeatable

maurodoglio:


moztelemetry/dataset.py
```diff
@@ -176,6 +176,9 @@ def records(self, sc, limit=None, sample=1, decode=None, summaries=None):
         if summaries and limit is None and sample != 1:
             if sample < 0 or sample > 1:
                 raise ValueError('sample must be between 0 and 1')
+            # We want this sample to be reproducible.
+            # See https://bugzilla.mozilla.org/show_bug.cgi?id=1318681
+            random.seed(42)
```
mreid-moz: Should we make this a parameter of the `records` function, defaulting to `42`? Then it can be overridden if one wants a different sample of the same size.

maurodoglio: I added a `seed` parameter to cover this use case as well as the case where we want to preserve the current behavior (no seed)

moztelemetry/dataset.py
```diff
@@ -176,8 +178,15 @@ def records(self, sc, limit=None, sample=1, decode=None, summaries=None):
         if summaries and limit is None and sample != 1:
             if sample < 0 or sample > 1:
                 raise ValueError('sample must be between 0 and 1')
-            summaries = random.sample(summaries,
-                                      int(len(summaries) * sample))
+            # We want this sample to be reproducible.
+            # See https://bugzilla.mozilla.org/show_bug.cgi?id=1318681
+            seed_state = random.getstate()
+            try:
+                random.seed(seed)
```
maurodoglio: exactly. by default it will behave exactly like now. but you can pass a seed to records() and it will produce constant samples

fbertsch: I'm of the opinion that the default behavior should be a reproducible random sample. It seems like an anti-pattern to return different subsets of data on different calls.

maurodoglio: The only reason why I put it as default is because it's backward-compatible. But I don't think anybody is basing their analysis on this subtle thing. I'll change it.

maurodoglio:
@fbertsch this is ready for another look

PR #130, opened 2017-03-30
Bug 1346308 - Support dataset prefixes with slashes

wlach:
So this seems almost too easy, but the unit tests (including a new one which would stress test whether this case is working correctly) are still passing. From what I can tell, when a prefix is specified we have a filter to make sure the data we fetch from S3 matches the prefix. This doesn't work well when our prefix has slashes in it (because it depends on being able to work against the last part of the path). However, I don't see why this is even necessary because we explicitly only get data matching the prefix from S3 (https://github.com/mozilla/python_moztelemetry/blob/06d26e6/moztelemetry/dataset.py#L128)

@mreid-moz @maurodoglio Could you tell me if I'm crazy?

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -148,10 +148,6 @@ class Dataset:
         clauses = copy(self.clauses)
         schema = self.schema
 
-        if self.prefix:
-            schema = ['prefix'] + schema
-            clauses['prefix'] = lambda x: x == self.prefix
-
         with futures.ProcessPoolExecutor(MAX_CONCURRENCY) as executor:
             scanned = self._scan(schema, [self.prefix], clauses, executor)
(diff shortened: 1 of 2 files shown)


mreid-moz:
I spent a bit of time puzzling through this, and I think you're right. Instead of creating a "fake" path component for the `prefix` and adding a clause to make sure it matches, your approach just skips ahead to the desired post-condition (as if we'd already done that for each component of the prefix).

Have you tried it on an actual cluster, running against S3 instead of the `InMemoryStore`? I am overly paranoid when it comes to changing this particular part of the system :)

wlach:
I managed to reproduce the unit test behaviour in an ipython notebook on atmo:

https://gist.github.com/wlach/1388af836a0ad0f8780d4918a1709383

PR #131, opened 2017-03-30
Fix hbase test fixtures when tests interrupted

wlach:
If you interrupt the hbase unit tests when they're running, that can leave
a table in a disabled-but-not-deleted state which the current fixture
can't handle correctly. Let's account for this case.


wlach:
@vitillo This fixes the issue we were just talking about.

PR #134, opened 2017-04-03
Experiment with creating a docker-based test environment

wlach:


wlach:
@mreid-moz @vitillo thoughts on this?

wlach:
Hmm, it seems as if coveralls doesn't work unless it's run just after the tests...

wlach:
Also, it occurs to me that we should probably make sure that the unit tests, when run locally, always use a fresh copy of the package/tests (e.g. via this procedure http://stackoverflow.com/a/23455537). It would be pretty confusing/annoying to update the code/tests locally and not see different results.

PR #136, opened 2017-04-04
Bug 1353196 - Fix malformed categorical histograms

fbertsch:


moztelemetry/histogram.py
```diff
@@ -133,7 +133,10 @@ def __init__(self, name, instance, revision=None, histograms_url=None,
             pd_index = labels + [CATEGORICAL_HISTOGRAM_SPILL_BUCKET_NAME]
             if isinstance(instance, dict):
                 # pandas clobbers values if instance is a dict with string labels
-                entries = {int(k): v for k, v in instance["values"].items()}
+                try:
+                    entries = {int(k): v for k, v in instance["values"].items()}
```
vitillo: Can we do the same also for non-categorical histograms?

fbertsch: Done and done.

fbertsch:
@vitillo RFAL

moztelemetry/histogram.py
vitillo: Genera comment not related to this patch: this code is pretty ugly, would be nice to refactor the initialization logic for the different histogram types in separate methods.

fbertsch: Agree wholeheartedly. The Histogram class as a whole could use some love.

PR #137, opened 2017-04-04
Specify happybase as a >= dependency, not a == dependency

wlach:
Since we are just a library that others include


wlach:
Just correcting an earlier mistake on my part

PR #139, opened 2017-04-06
Abort (and fail) if any step in the tests fail

wlach:


wlach:
This should fix the unit test failures (and make sure we pick up on them in the future), at least. Still trying to figure out what's gone wrong in production.

PR #142, opened 2017-04-11
Better testing of end-to-end heka parsing

wlach:


tests/heka/test_message_parser.py
```diff
@@ -85,3 +96,24 @@ def test_invalid_utf8(data_dir):
...
+    mock_parse_json = MagicMock(name='_parse_json',
+                                wraps=_parse_json)
+    import moztelemetry.heka.message_parser
+    monkeypatch.setattr(moztelemetry.heka.message_parser, '_parse_json', mock_parse_json)
+
+    filename = "{}/test_telemetry_gzip.heka".format(data_dir)
+    with open(filename, "rb") as o:
+        heka_message = message_parser.parse_heka_message(streaming_gzip_wrapper(o)).next()
+
+    # should only have parsed json *once* (json fields of the message should
```
vitillo: Where is that single call made?

tests/test_spark.py
```diff
@@ -156,3 +157,30 @@ def test_get_pings_none_filter(test_store, dummy_pool_executor,
     pings = get_pings(spark_context, app='*')
 
     assert sorted(pings.collect()) == ['value1', 'value2']
+
+
+def test_get_heka_pings(data_dir, test_store, dummy_pool_executor, monkeypatch,
+                        spark_context):
+    # this monkeypatching doesn't actually work, because the code is run in a
```
vitillo: @maurodoglio might have some ideas about this

tests/heka/test_message_parser.py
```diff
@@ -85,3 +96,24 @@ def test_invalid_utf8(data_dir):
...
+    # be parsed lazily)
+    assert mock_parse_json.call_count == 1
+
+    # deep copying the heka message should cause the lazily evaluated fields
+    # to be evaluated
+    new_heka_message = copy.deepcopy(heka_message)
+    assert mock_parse_json.call_count == 20
```
vitillo: Are we sure it should be exactly 20? Please define magic constants.

tests/heka/test_message_parser.py
wlach: At the very beginning of `_parse_heka_record` in `message_parser.py`. Since the payload forms the skeleton of the dictionary, I suspect there isn't any way to lazily process it.

tests/heka/test_message_parser.py
```diff
@@ -4,61 +4,72 @@
...
-            assert expected_counts[t] == count
 
+@pytest.mark.parametrize("heka_format,expected_count", [
+    ("plain", 10),
+    ("snappy", 0),
+    ("mixed", 5),
+    ("gzip", 10),
+    ("gzip_mixed", 5)
+])
+def test_unpack_nosnappy(data_dir, heka_format, expected_count):
```
wlach: Ok, consolidated all of this into a single parametrized test. I think it's easier to see what's going on now...

tests/heka/test_message_parser.py
```diff
@@ -4,61 +4,72 @@
...
+    count = 0
+    filename = "{}/test_{}.heka".format(data_dir, heka_format)
+    with open(filename, "rb") as o:
+        if "gzip" in heka_format:
+            o = streaming_gzip_wrapper(o)
+        try:
+            for r, b in message_parser.unpack(o, try_snappy=False):
+                count += 1
```
wlach: No, fixed that.

tests/heka/test_message_parser.py
wlach: I added a comment.

tests/test_spark.py
wlach: I will disable the test for now.

whd:
I started taking a look at this but the PR changed substantially while I was reviewing it. I will take another look tomorrow.

wlach:
@whd sorry about that, I was addressing @vitillo's comments.

wlach:
Looking at the coverage stats, the "improvement" seems to be that we are now testing that we actually are able to deserialize the json blobs inside a telemetry heka message.

whd:
It appears there are some [flake8 issues](https://travis-ci.org/mozilla/python_moztelemetry/jobs/221816051) with the current PR.

wlach:
Gah, foiled by my only newly-created tests! At least this shows they're working/useful. Sorry about the confusion, will update...

wlach:
@whd should be ok now

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,105 @@
...
-        with open(filename, "rb") as o:
-            if "gzip" in t:
-                o = streaming_gzip_wrapper(o)
-            try:
-                for r, b in message_parser.unpack(o, strict=True, try_snappy=False):
-                    count += 1
-            except Exception:
-                threw = True
-        assert expected_exceptions[t] == threw
+    # enable this to regenerate the expected json representation of the ping
```
whd: I'm not sure I like adding the generated JSON reference blobs to the repo, as the real reference blobs are those submitted by the client. I think the better thing to do would be to use a raw ping from the client (from e.g. about:telemetry) as the reference, and re-derive the legacy heka decoded and current infra decoded versions from that. This it turns out is not trivial to do, esp. for the deprecated infra, but it is possible. The test as written assumes that our code is working correctly now, so as a regression test it is acceptable. I don't think it's necessarily worth the effort involved to recreate our test data using an actual reference ping, but I thought I would call it out.

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,105 @@
...
+    if False:
+        with open(filename, "rb") as f:
+            output = []
```
whd: I'm guessing that you're generating a list because `parse_heka_message` returns an iterator, even though there is only one item per reference file. I don't feel strongly about this, but the structure seems odd. For style it might be good to match other single-message-per-file tests like the invalid UTF8 one, which iterates over a list of one implicitly (or change the UTF8 test to match this one).

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,105 @@
...
+            if "gzip" in heka_format:
+                f = streaming_gzip_wrapper(f)
+            for r in message_parser.parse_heka_message(f):
+                output.append(copy.deepcopy(r))
```
whd: It was not immediately obvious to me why the deep copies here and below are necessary, but there is commentary further down in the lazy parsing tests that explains it.

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,105 @@
...
+                                wraps=_parse_json)
+    import moztelemetry.heka.message_parser
+    monkeypatch.setattr(moztelemetry.heka.message_parser, '_parse_json', mock_parse_json)
+
+    filename = "{}/test_telemetry_gzip.heka".format(data_dir)
+    with open(filename, "rb") as o:
+        heka_message = message_parser.parse_heka_message(streaming_gzip_wrapper(o)).next()
+
+    # should only have parsed json *once* to get the payload (json/dictionary
+    # fields of the message should be parsed lazily)
```
whd: The comments here and below are helpful, but it might be nice to have a preliminary comment that explicitly mentions that the test input data has 20 fields, 19 of which should be parsed lazily. This can be inferred from the final assertion's comment, but it is not explicit.

tests/test_spark.py
```diff
@@ -156,3 +158,21 @@ def test_get_pings_none_filter(test_store, dummy_pool_executor,
...
+@pytest.mark.skip("monkeypatching doesn't persist across spark context")
+def test_get_heka_pings(data_dir, test_store, dummy_pool_executor, monkeypatch,
+                        spark_context):
+    from moztelemetry.heka.message_parser import _parse_json
+    mock_parse_json = MagicMock(name='_parse_json',
+                                wraps=_parse_json)
+    import moztelemetry.heka.message_parser
+    monkeypatch.setattr(moztelemetry.heka.message_parser, '_parse_json', mock_parse_json)
+
+    heka_content = open("{}/test_telemetry_snappy.heka".format(data_dir), "rb").read()
```
whd: This test appears to be disabled, but `heka_content` here should I think be derived from test_telemetry_gzip.heka since that is the current standard format (snappy is the legacy one). I assume from this test passing that the number of json  fields in the snappy heka message  is the same as the gzip one, which is not necessarily obvious as they are different messages.

tests/heka/test_message_parser.py
wlach: Yeah, the idea is to catch inadvertent regressions in the parser. My experience is that "real" test data is not that valuable for unit tests that test this sort of etl code -- better to have a well-formatted set of test data that covers as many (edge) cases as possible.

tests/heka/test_message_parser.py
wlach: Agreed that this is a bit confusing. I'll change it so we just get one element from the generator and use that.

tests/heka/test_message_parser.py
wlach: I'll add some one line comments. In general I'm not crazy about this hack but not sure of a better way of doing this in the short term.

tests/heka/test_message_parser.py
wlach: Ok, will add one.

tests/test_spark.py
wlach: Ok, will change.

wlach:
@whd updated per your comments; I'd be happy to test the ujson -> ordinary json fallback if you can give me an example of where that should happen (google didn't give me any)

tests/test_spark.py
whd: I may be reading the UI incorrectly, but this line appears unchanged in the latest version of the PR.

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,104 @@
...
+
+
+def test_lazy_parsing(data_dir, monkeypatch):
+    from moztelemetry.heka.message_parser import _parse_json
+    mock_parse_json = MagicMock(name='_parse_json',
+                                wraps=_parse_json)
+    import moztelemetry.heka.message_parser
+    monkeypatch.setattr(moztelemetry.heka.message_parser, '_parse_json', mock_parse_json)
+
+    # this heka message has 19 json fields, none of which should be parsed on
```
whd: Technically there are 20 json fields in the new data. There's a long, sordid explanation here, but in short if fields here means "protobuf fields", there are 20 in both the new and old data. If you're referring specifically to [our protobuf schema's Fields](https://github.com/mozilla-services/heka/blob/dev/message/message.proto#L22-L38), there are 19 in the old infra data and 20 in the new infra data (the old infra put the special payload in the top-level `Payload` field). Either way, there are 20 fields in the new infra data, of which one is parsed specially. This does mean we iterate over fields twice when decoding new data ([here](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/heka/message_parser.py#L32) and [here](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/heka/message_parser.py#L52)), which is inefficient but makes the code much easier to implement (I originally incorrectly implemented it using one pass).

wlach: Ah yes, you're right. Sorry, I misread the code. I'll update the comments appropriately.

wlach:
I forgot to commit the changes to the spark changes before pushing, updated now. Also added a test for verifying that the json fallback works (and verified it catches the problem if the safety-valve is removed)

tests/test_spark.py
```diff
@@ -156,3 +158,21 @@ def test_get_pings_none_filter(test_store, dummy_pool_executor,
...
+@pytest.mark.skip("monkeypatching doesn't persist across spark context")
+def test_get_heka_pings(data_dir, test_store, dummy_pool_executor, monkeypatch,
+                        spark_context):
+    from moztelemetry.heka.message_parser import _parse_json
+    mock_parse_json = MagicMock(name='_parse_json',
+                                wraps=_parse_json)
+    import moztelemetry.heka.message_parser
+    monkeypatch.setattr(moztelemetry.heka.message_parser, '_parse_json', mock_parse_json)
+
+    heka_content = open("{}/test_telemetry_gzip.heka".format(data_dir), "rb").read()
```
whd: This test is disabled, but this data may need to be gunzipped before being uploaded, or have its content-type set to gzip (not sure how the upload_ping or test_store stuff are implemented).

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,113 @@
...
 
 def test_invalid_utf8(data_dir):
     filename = "{}/test_invalid_utf8.heka".format(data_dir)
     with open(filename, "rb") as o:
         for r in message_parser.parse_heka_message(o):
             assert(u'\ufffd' in r['info']['adapterDescription'])
+
+
+def test_json_fallback():
+    # ujson can't handle values which are too large ()
```
whd: There's an extra () here.

It also might be good to explicitly assert the ujson failure from the issue. That way, if the assertion ever fails we can remove the json fallback code. The link to the issue is probably sufficient however.

tests/test_spark.py
wlach: it seemed to work when I enabled the test, although of course the thing I was trying to test didn't. my preference would be just to leave this as-is for now; we can revisit if/when we decide to actually fix the test.

tests/heka/test_message_parser.py
wlach: updated

whd:
LGTM, over to @vitillo for additional review.

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,121 @@
...
-    with open(filename, "rb") as o:
-        for r in message_parser.parse_heka_message(o):
-            assert set(r.keys()) == top_keys
-            assert set(r["payload"].keys()) == payload_keys
+    assert count == expected_count
+    assert threw_exception == expected_exception
+
+
+@pytest.mark.parametrize("heka_format", ["snappy", "gzip"])
+def test_telemetry(data_dir, heka_format):
```
maurodoglio: Let's rename this to something less generic, like `test_parse_heka_message`

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,121 @@
...
+    with pytest.raises(ValueError):
+        ujson.loads(str(TOO_BIG))
+
+    # now assert that the fallback to standard json works
+    from moztelemetry.heka.message_parser import _parse_json
+    assert TOO_BIG == _parse_json(str(TOO_BIG))
+
+
+def test_lazy_parsing(data_dir, monkeypatch):
+    from moztelemetry.heka.message_parser import _parse_json
```
maurodoglio: Could you combine these imports into one? importing `message_parser` should make it work for both MagicMock (message_parser._parse_json) and monkeypatch

tests/test_spark.py
```diff
@@ -156,3 +158,21 @@ def test_get_pings_none_filter(test_store, dummy_pool_executor,
     pings = get_pings(spark_context, app='*')
 
     assert sorted(pings.collect()) == ['value1', 'value2']
+
+
+@pytest.mark.skip("monkeypatching doesn't persist across spark context")
+def test_get_heka_pings(data_dir, test_store, dummy_pool_executor, monkeypatch,
+                        spark_context):
+    from moztelemetry.heka.message_parser import _parse_json
```
maurodoglio: Same here as above

tests/test_spark.py
maurodoglio: Let's use Dataset for this test, get_pings is kind of deprecated and will be removed one day.

maurodoglio: If you want to run an end-to-end test like this you should probably use https://github.com/spulec/moto#stand-alone-server-mode. To make it work with Dataset, you need to modify S3Store so that you can pass a custom `endpoint_url` to `boto3.resource` and `boto3.client`.

tests/heka/test_message_parser.py
```diff
@@ -1,87 +1,121 @@
...
+    assert threw_exception == expected_exception
+
+
+@pytest.mark.parametrize("heka_format", ["snappy", "gzip"])
+def test_telemetry(data_dir, heka_format):
+    filename = "{}/test_telemetry_{}.heka".format(data_dir, heka_format)
+    reference_filename = filename + '.json'
+
+    # enable this to regenerate the expected json representation of the ping
+    if False:
```
wlach: I feel like we want generating the file to be a conscious choice, so I think I'm going to leave this as-is for now.

wlach:
@maurodoglio do you know how to get around the fact that monkeypatching doesn't persist across spark contexts?

maurodoglio:
I guess you mean across local python <-> pyspark worker? That's not possible afaik. The only thing you can do is to share serializable objects. That's why I created InMemoryStore, which is just a way to have shareable S3Store-like object.

wlach:
Ok, should we just take out the already-disabled test in `test_spark.py` then, since we have no way of knowing what happens inside the spark context? Or is there a way to make it work?

maurodoglio:
My [comment](https://github.com/mozilla/python_moztelemetry/pull/142#discussion_r112463904) above was suggesting to use a standalone moto server to run the test using a regular S3Store. That would allow you to upload a sample gzipped ping (using S3Store), use Dataset to fetch it and then inspect the resulting payload. That's the only idea I could come up with 😞

wlach:
Yeah, I don't think that's going to test the thing we care about then -- I think I'll just drop this test.

maurodoglio:
SGTM

PR #149, opened 2017-05-09
Fix running tests with arguments

wlach:


wlach:
bash is so finnicky, the first pass didn't work. trying again

PR #150, opened 2017-05-10
Bug 1318326 - Fix `get_pings_properties` for keyedHistograms

acmiyaguchi:
[Bug 1318326 - |get_pings_properties| does not return all the available keyedHistograms](https://bugzilla.mozilla.org/show_bug.cgi?id=1318326)

This PR introduces failing tests that demonstrate the current behavior of `get_pings_properties` when retrieving keyedHistograms in pings that contain child processes. This is solved by defaulting to empty dictionaries to avoid `KeyError` exceptions.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/150)

### DIFF
--- a/moztelemetry/spark.py
+++ b/moztelemetry/spark.py
@@ -29,6 +29,24 @@ except:
 _sources = None
 
 
+class PingCursor(dict):
+    """ A subclassed dictionary that defaults to a new instance of
+    itself when missing a key. See http://stackoverflow.com/a/19829714
+    for more details.
+    """
+
+    def __getitem__(self, key):
+        value = dict.__getitem__(self, key)
+        if isinstance(value, dict):
+            # cast all subdicts to PingCursor
+            value = type(self)(value)
+        return value
+
+    def __missing__(self, key):
+        value = self[key] = type(self)()
+        return value
+
+
 def get_pings(sc, app=None, build_id=None, channel=None, doc_type='saved_session',
               fraction=1.0, schema=None, source_name='telemetry', source_version='4',
               submission_date=None, version=None):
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
The updated PR fixes test failures due to bug 1363934. This also adds one more edge case that wasn't caught in the initial PR. 

The updated notebook that verifies the results can be found [here](https://gist.github.com/acmiyaguchi/0f222374f9f622b4fc3415987811ffb5).

Dexterp37:
This looks good to me, but @fbertsch might be a more appropriate reviewer as I haven't touched this code in ages.

moztelemetry/spark.py
```diff
@@ -182,16 +182,20 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
                                            .get("content", {}) \
                                            .get("keyedHistograms", None)
                         if content_kh is not None:
-                            kh_keys.update(content_kh[path[1]].keys())
+                            kh_keys.update(content_kh.get(path[1], {}).keys())
                         else:
                             for payload in cursor.get("childPayloads", []):
-                                kh_keys.update(payload["keyedHistograms"][path[1]].keys())
+                                kh_keys.update(
```
fbertsch: Please replace all uses of update with `kh_keys |= someDict.viewkeys()`; it makes it more apparent this is a set and not a dict.

moztelemetry/spark.py
```diff
@@ -182,16 +182,20 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
                                            .get("content", {}) \
```
fbertsch: Mauro and I were discussing this and think that using a safe dict as the cursor might be a better (and more readable) solution. See [What is the best way to implement nested dictionaries in Python?](http://stackoverflow.com/questions/635483/what-is-the-best-way-to-implement-nested-dictionaries/19829714#19829714)

Then we can replace all `.get(elem)` with just `[elem]`

moztelemetry/spark.py
```diff
@@ -182,16 +182,20 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
...
+                                        .get(path[1], {})
+                                        .keys())
 
                         gpu_kh = cursor.get("processes", {}) \
                                        .get("gpu", {}) \
                                        .get("keyedHistograms", None)
                         if gpu_kh is not None:
-                            kh_keys.update(gpu_kh[path[1]].keys())
+                            kh_keys.update(gpu_kh.get(path[1], {}).keys())
                 except:
```
acmiyaguchi: I'm all for dropping this exception, it seems to be just catching KeyErrors.

acmiyaguchi:
@fbertsch This is ready for another look. I've implemented the safe dictionary and fixed any places where it could have changed the logic. I looked over the Histogram implementation, but I don't think the new KeyError-less behavior affects anything there.

moztelemetry/spark.py
```diff
@@ -162,39 +180,34 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
...
-                            kh_keys.update(gpu_kh[path[1]].keys())
-                except:
-                    result[property_name] = None
-                    continue
+                # These are returned as a subdict mapped from the
+                # property_name.
+                kh_keys = cursor["keyedHistograms"][path[1]].viewkeys()
+                if isinstance(cursor, dict):
+                    content_kh = cursor["processes"]["content"]["keyedHistograms"]
+                    if content_kh:
```
fbertsch: We don't need these ifs anymore, do we? The safe cursor will return an empty `dict`, where `content_kh[path[1]].viewkeys()` will return an empty set. So you could reduce this to:

`kh_keys |= cursor["processes"]["content"]["keyedHistograms"][path[1]].viewkeys()`

Same with `gpu_kh`.

moztelemetry/spark.py
```diff
@@ -162,39 +180,34 @@ def _get_ping_properties(ping, paths, only_median, with_processes,
...
-                        if gpu_kh is not None:
-                            kh_keys.update(gpu_kh[path[1]].keys())
-                except:
-                    result[property_name] = None
-                    continue
+                # These are returned as a subdict mapped from the
+                # property_name.
+                kh_keys = cursor["keyedHistograms"][path[1]].viewkeys()
+                if isinstance(cursor, dict):
+                    content_kh = cursor["processes"]["content"]["keyedHistograms"]
```
fbertsch: It occurs to me that this will fail if `cursor["processes"]`, `cursor["processes"]["content"]`, etc. are present but not `dict`s. Do you know if that schema is mandated upstream?

acmiyaguchi: This should be mandated by the ingestion schema's. Here are the relevant blobs from the main ping schema that defines `processes` and `content` as objects.

https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/main/main.4.schema.json#L556-L567
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/main/main.4.schema.json#L875-L876

moztelemetry/spark.py
acmiyaguchi: That's true, I'll make comment of why this was structured this way.

acmiyaguchi:
@fbertsch this is ready for another look

PR #151, opened 2017-05-11
Bug 1363934 - Update moztelemetry's ages-old histogram_tools, fixing tests.

chutten:
It's been a while since moztelemetry's copy of histogram_tools was updated so
there are two things that need to be addressed.

1) We need to handle that Categorical histograms have a default of 50 n_buckets
So truncate values as moztelemetry.Histograms are constructed

2) record_in_processes is required in histogram definitions
So use more recent versions of histogram definitions that comply

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/151)


chutten:
They are precisely the versions in mozilla-central, so yes.

PR #157, opened 2017-06-19
Update histogram_tools.py

Dexterp37:
The new version includes a check to ignore unknown keys in the histograms file when ran from the pipeline.
This also updates the shared_telemetry_utils.py copy.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/python_moztelemetry/157)


Dexterp37:
@fbertsch all the tests pass, is there anything else that I can do (other than watching once this hits the pipeline) to mitigate the risk of breaking stuff?

We should be fairly solid on this one, but you know.. you never know :)

PR #160, opened 2017-06-28
Bug 1376028 - Stop using Lazy json evaluation

mreid-moz:
This has potential performance implications. We should do some testing to see how it performs when we just parse all the json.


mreid-moz:
@fbertsch flake8 is slain. ready for a look.

fbertsch:
@mreid-moz it took 1271s on the old code, and 842s on the new (without lazy parsing). I don't think this is going to have a significant effect, if any at all.

mreid-moz:
What the heck? How can it be faster? Let's verify that we've actually fixed the issue you demonstrated in the notebook in the bug.

PR #161, opened 2017-06-28
Bug 1306049 - Sanitize string dimensions passed to "where"

mreid-moz:


moztelemetry/dataset.py
```diff
@@ -154,6 +156,18 @@ def _apply_selection(self, json_obj):
         return dict((name, path.search(json_obj))
                     for name, path in self.selection_compiled.items())
 
+    def _sanitize_dim(self, v):
+        """For String conditions, we should pre-sanitize so that users of
```
maurodoglio: Could you format this docstring to follow the pep [257 conventions](https://www.python.org/dev/peps/pep-0257/#id17)?

moztelemetry/dataset.py
```diff
@@ -154,6 +156,18 @@ def _apply_selection(self, json_obj):
...
+        the `where` function do not need to know about the nuances of how
+        S3 dimensions are sanitized.
+
+        See https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/io_modules/moz_telemetry/s3.lua#L167
+
+        :param v: a string value that should be sanitized.
+        """
+        if v is not None:
```
maurodoglio: I don't think this condition is needed

moztelemetry/dataset.py
```diff
@@ -169,6 +183,8 @@ def where(self, **kwargs):
                 raise Exception('The dimension {} doesn\'t exist'.format(dimension))
             if isfunction(condition) or isinstance(condition, functools.partial):
                 clauses[dimension] = condition
+            elif isinstance(condition, basestring):
```
maurodoglio: The value that this condition is checked against is a string, so we can either raise an exception if this is not the case or convert condition to a string. I think the latter solution would make more sense but I don't have a strong opinion about it.

mreid-moz:
@maurodoglio RFAL

PR #170, opened 2017-08-09
Bug 1380438 - Make scalar names case-insensitive

fbertsch:
We are guaranteed that no two scalars in scalars.yaml will share
a case-insensitive name. Currently, the Scalar class fails to
return a scalar when given a proper scalar name that is not
cased correctly. This change matches any case.

### DIFF
--- a/moztelemetry/scalar.py
+++ b/moztelemetry/scalar.py
@@ -23,6 +23,10 @@ REVISIONS = {'nightly': 'https://hg.mozilla.org/mozilla-central/rev/tip',
              'release': 'https://hg.mozilla.org/releases/mozilla-release/rev/tip'}
 
 
+class MissingScalarError(KeyError):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


moztelemetry/scalar.py
```diff
@@ -110,4 +110,7 @@ def _get_scalar_definition(url, metric):
         else:
             definitions = Scalar._definition_cache[url]
 
+        if metric.lower() not in definitions:
+            raise KeyError("Definition not found for {}".format(metric))
```
maurodoglio: What about using a custom exception? That way calling code can except this specific error and not a generic KeyError.

maurodoglio:
Also, Travis is unhappy because of the duplicate test name, could you please change it?

fbertsch:
@maurodoglio good suggestions! Thanks for the review!

PR #171, opened 2017-08-10
Update protobuf version

maurodoglio:
See the commit messages for a description


moztelemetry/histogram.py
```diff
@@ -59,14 +59,7 @@ def cached_linear_buckets(*args, **kwargs):
...
     if cached is None:
-        definition = requests.get(url).content
-
-        # see bug 920169
-        definition = definition.replace('"JS::gcreason::NUM_TELEMETRY_REASONS"', "101")
-        definition = definition.replace('"mozilla::StartupTimeline::MAX_EVENT_ID"', "12")
-        definition = definition.replace('"80 + 1"', "81")
-
-        parsed = json.loads(definition)
+        parsed = requests.get(url).json()
```
mreid-moz: This is a nice cleanup! I worry a little that this function could be used to load an old histograms URL, which would break without these special cases.

This URL is exposed in the public API in `spark.py` (in `get_pings_properties`), so a caller could pass in the URL for Histograms.json for an old Firefox version (supposing they were analyzing data for recent data submitted by an old version).

I think the risk is quite low, but if that happened, it would make for quite an unpleasant debugging experience.

maurodoglio: nevermind, I'm gonna drop this commit

PR #174, opened 2017-09-07
Add Mann-Whitney U test methods (bug 1395571)

robhudson:
TODO:

- [x] Also return p-values, similar to scipy

### DIFF
--- a/moztelemetry/stats.py
+++ b/moztelemetry/stats.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from __future__ import division
+
+import math
+from collections import namedtuple
+
+
+def _rank(sample):
+    """
+    Assign numeric ranks to all values in the sample.
+
+    The ranks begin with 1 for the smallest value. When there are groups of
+    tied values, assign a rank equal to the midpoint of unadjusted rankings.
+
+    E.g.::
+
+        >>> rank({3: 1, 5: 4, 9: 1})
+        {3: 1.0, 5: 3.5, 9: 6.0}
+
+    """
+    rank = 1
+    ranks = {}
+
+    for k in sorted(sample.keys()):
+        n = sample[k]
+        ranks[k] = rank + (n - 1) / 2
+        rank += n
+
+    return ranks
+
+
+def _tie_correct(sample):
+    """
+    Returns the tie correction value for U.
+
+    See: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.tiecorrect.html
+
+    """
+    tc = 0
+    n = sum(sample.values())
+
+    if n < 2:
+        return 1.0  # Avoid a ``ZeroDivisionError``.
+
+    for k in sorted(sample.keys()):
+        tc += math.pow(sample[k], 3) - sample[k]
+    tc = 1 - tc / (math.pow(n, 3) - n)
+
+    return tc
+
+
+def ndtr(a):
+    """
+    Returns the area under the Gaussian probability density function,
+    integrated from minus infinity to x.
+
+    See: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ndtr.html#scipy.special.ndtr
+
+    """
+    sqrth = math.sqrt(2) / 2
+    x = float(a) * sqrth
+    z = abs(x)
+    if z < sqrth:
+        y = 0.5 + 0.5 * math.erf(x)
+    else:
+        y = 0.5 * math.erfc(z)
+        if x > 0:
+            y = 1 - y
+    return y
+
+
+mwu_result = namedtuple('Mann_Whitney_U', ('u', 'p'))
+
+
+def mann_whitney_u(sample1, sample2, use_continuity=True):
+    """
+    Computes the Mann-Whitney rank test on both samples.
+
+    Each sample is expected to be of the form::
+
+        {1: 5, 2: 20, 3: 12, ...}
+
+    Returns a named tuple with:
+        ``u`` equal to min(U for sample1, U for sample2), and
+        ``p`` equal to the p-value.
+
+    """
+    # Merge dictionaries, adding values if keys match.
+    sample = sample1.copy()
+    for k, v in sample2.items():
+        sample[k] = sample.get(k, 0) + v
+
+    # Create a ranking dictionary using same keys for lookups.
+    ranks = _rank(sample)
+
+    sum_of_ranks = sum([sample1[k] * ranks[k] for k, v in sample1.items()])
+    n1 = sum(sample1.values())
+    n2 = sum(sample2.values())
+
+    # Calculate Mann-Whitney U for both samples.
+    u1 = sum_of_ranks - (n1 * (n1 + 1)) / 2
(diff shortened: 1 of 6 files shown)


moztelemetry/stats.py
```diff
@@ -0,0 +1,45 @@
...
+    # Create a ranking dictionary using same keys for lookups.
+    rank = 1
+    ranks = {}
+    for k in sorted(sample.keys()):
+        n = sample[k]
+        ranks[k] = rank + (n - 1) / 2
+        rank += n
+
+    # Calculate Mann-Whitney U for both samples.
+    sum_of_ranks1 = sum([sample1[k] * ranks[k] for k, v in sample1.items()])
```
fbertsch: Repeated code here. Instead of doing the calculation for `U2`, you could instead use the fact that `U1 + U2 == n1 * n2`.

tests/test_stats.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+from moztelemetry import stats
+
+
+def test_mann_whitney_u():
+    SAMPLE1 = {1: 5, 2: 1, 3: 4, 4: 2, 5: 3}
+    SAMPLE2 = {1: 1, 2: 5, 3: 2, 4: 4, 5: 3}
+
+    # Basic test.
+    assert stats.mann_whitney_u(SAMPLE1, SAMPLE2) == 94.5
```
fbertsch: I think you should import scipy's mannwhitneyu test and compare its test statistic to yours. Could even have comparisons on some generated histograms.

robhudson:
It was noted on IRC that I can borrow the method for calculating p-values from the lua code here:
https://github.com/mozilla-services/lua_sandbox_extensions/blob/d5d65adbe4e075825aa99f5abee01ffff93bcced/lsb/modules/lsb/stats.lua#L311-L319

tests/test_stats.py
robhudson: This project doesn't currently depend on scipy. Is that an ok requirement to add?

robhudson: I tried to simply add it to the setup.py similar to numpy but it requires a fortran compiler and failed. It'll be a bit more complicated to add this it seems?

moztelemetry/stats.py
```diff
@@ -0,0 +1,92 @@
...
+    n1 = sum(sample1.values())
+    n2 = sum(sample2.values())
+
+    u1 = sum_of_ranks - (n1 * (n1 + 1)) / 2
+    u2 = n1 * n2 - u1
+
+    tie_correction = tie_correct(sample)
+    sd_u = math.sqrt(tie_correction * n1 * n2 * (n1 + n2 + 1) / 12.0)
+    mean_rank = n1 * n2 / 2.0 + 0.5 * use_continuity
+    z = abs((max(u1, u2) - mean_rank) / sd_u)
```
fbertsch: This should be the `min(u1, u2)`, [see here](http://psych.unl.edu/psycrs/handcomp/hcmann.PDF)

moztelemetry/stats.py
```diff
@@ -0,0 +1,92 @@
...
+        return 1.0  # Avoid a ``ZeroDivisionError``.
+
+    for k in sorted(sample.keys()):
+        tc += math.pow(sample[k], 3) - sample[k]
+    tc = 1 - tc / (math.pow(n, 3) - n)
+
+    return tc
+
+
+def ndtr(v):
```
fbertsch: `ndtr` takes the integral from `-infinity` to `v` over the gaussian distribution. This is why the input below is `-abs(z)`. Can we do two things here:

- [ ] Add tests that this function returns the same value as [scipy ndtr](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ndtr.html#scipy.special.ndtr)

- [ ] Document what this function is doing, with a reference to the above link

I do like calculating this separately, rather than importing scipy.

tests/test_stats.py
```diff
@@ -0,0 +1,146 @@
...
+# encoding: utf-8
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+This module implements test coverage for the stats functions in stats.py.
+"""
+import itertools
```
fbertsch: Can we add more tests, with some different distributions? We could add, for example:

- Normally distributed histograms
- Skewed distributions
- Uniformly distributed histograms

And combinations therein. We can create them randomly, and check that our result is within scipy's by some value.

moztelemetry/stats.py
robhudson: In step 10 they're using U=32 to compute `z`, which is the `max(Ua, Ub)` in the example.

moztelemetry/stats.py
```diff
@@ -0,0 +1,92 @@
...
+def mann_whitney_u(sample1, sample2, use_continuity=True):
+    """
+    Computes the Mann-Whitney rank test on both samples.
+
+    Each sample is expected to be of the form::
+
+        {1: 5, 2: 20, 3: 12, ...}
+
+    Returns the U statistic, equal to min(U for sample1, U for sample2).
+
```
robhudson: Update comments to match code.

moztelemetry/stats.py
```diff
@@ -0,0 +1,92 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from __future__ import division
+
+import math
+from collections import namedtuple
+
+
+def rank(sample):
```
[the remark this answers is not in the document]
robhudson: It will return doubles since I'm importing division from the `__future__`, which will be more Python 3 if this project ever moves to it. W/o that import you're correct (and living in the past).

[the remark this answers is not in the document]
fbertsch: heh, good point! I've gotta switch my local default to 3.6...

robhudson:
Updated with requested changes.

moztelemetry/stats.py
```diff
@@ -0,0 +1,120 @@
...
+def ndtr(a):
+    """
+    Returns the area under the Gaussian probability density function,
+    integrated from minus infinity to x.
+
+    See: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ndtr.html#scipy.special.ndtr
+
+    """
+    try:
+        a = float(a)
```
fbertsch: It is probably more pythonic to let this exception be raised. Perhaps just line 72 should be `float(a) * sqrth`. Alternatively, we could just remove validation entirely.

tests/test_stats.py
```diff
@@ -0,0 +1,109 @@
...
+
+    assert round(stats.ndtr(0), 6) == round(scipy.special.ndtr(0), 6)
+    assert round(stats.ndtr(1), 6) == round(scipy.special.ndtr(1), 6)
+    assert round(stats.ndtr(1.5), 6) == round(scipy.special.ndtr(1.5), 6)
+    assert round(stats.ndtr(2), 6) == round(scipy.special.ndtr(2), 6)
+    assert round(stats.ndtr(3), 6) == round(scipy.special.ndtr(3), 6)
+
+
+def test_mann_whitney_u():
+    for sample in ('normalized', 'uniform', 'skewed'):
```
fbertsch: These tests are close to what I was thinking, but all of these are comparing the same distribution. I would change one of the distributions (for each of: normalized, uniform, and skewed) to be different (different mean, different std dev, different number of samples, etc.), and in addition add tests that compare e.g. normalized to uniform, uniform to skewed, etc.

PR #175, opened 2017-09-08
Bug 1397955 - get_pings_properties: Handle boolean scalars with false values

mnoorenberghe:
r? @acmiyaguchi


mnoorenberghe:
> This looks good to me. Did you run into any interesting failures when an empty Cursor was passed Histogram?

I'm not sure what you mean. The `len` check was to handle the empty cursor case. I didn't test any behaviour with histograms.

PR #177, opened 2017-09-28
Add a utility function for reading main_summary.

mreid-moz:
This supports merging schemas for only a subset of partitions.

Test data was generated using [this script](https://gist.github.com/mreid-moz/14f177f461486692af945176c1ede87d).


mreid-moz:
The build failure appears to be unrelated to these changes... maybe [this problem](https://stackoverflow.com/questions/46450912/unable-to-execute-x86-64-conda-cos6-linux-gnu-gcc-no-such-file-or-directory)?

mreid-moz:
Ok, tests are finally passing.

I removed the manual Spark installation, instead using the `pyspark` available from pip. I'm not sure exactly what implications this has when deploying to real clusters, so tagging Mauro for review as well.

I had to add a temporary workaround for the conda issue in my previous comment, I'll file a follow-up issue to fix that when (if?) this lands.

setup.py
```diff
@@ -17,7 +17,7 @@
     package_dir={'moztelemetry': 'moztelemetry'},
     install_requires=['boto', 'boto3', 'ujson', 'requests', 'protobuf',
                       'expiringdict', 'functools32', 'futures', 'py4j',
-                      'pandas>=0.14.1', 'numpy>=1.8.2', 'findspark',
+                      'pandas>=0.14.1', 'numpy>=1.8.2', 'pyspark',
```
fbertsch: Hmm, is this actually required? seems to me the tests would add this due to `environment.yaml`.

mreid-moz: Let's find out :)

mreid-moz:
Looks like it's safe to leave it out.

PR #178, opened 2017-10-18
Bug 1407406 - Dataset.select preserve dataset properties

maurodoglio:


wlach:
looks like travis is failing for reasons unrelated to this change.

wlach:
Would it be possible to add a unit test for this case? (can be in a seperate pr if you like)

maurodoglio:
Isn't [this unittest](https://github.com/mozilla/python_moztelemetry/pull/178/files#diff-ecf5283ea4e3b0739e1390fb223ada37) enough? What test case do you have in mind?

tests/test_dataset.py
```diff
@@ -60,6 +60,14 @@ def test_select():
     }
 
 
+def test_select_keep_clauses():
+    """Test that calling select doesn't discard filter clauses"""
+    dataset_before = Dataset('test-bucket', ['dim1', 'dim2']).where(dim1=True)
+    dataset_after = dataset_before.select('field1', 'field2')
+
+    assert dataset_before.clauses == dataset_after.clauses
```
wlach: Could you verify that we keep the other attributes (bucket, schema, store) as well?

wlach:
Argh, sorry, somehow missed the test the first time around. Might be nice to have some extra checks, LGTM otherwise.

maurodoglio:
@wlach the issue with the travis run is due to a recent change in moto that now requires a specific version of a library not available anymore on pypi. This is the upstream change that introduced the regression https://github.com/spulec/moto/commit/49ddb500a839daaee8b916fb0af365825d66467d

wlach:
Yup, I suspected it was something silly like that. :/

PR #185, opened 2017-11-02
Remove Coveralls

fbertsch:
Coveralls is currently making it impossible to merge PRs, since
they are not having the coveralls status reported. Admins also
cannot override this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ WORKDIR /python_moztelemetry
 
 # we need to explicitly install pytest and dependencies so spark
 # can pick them up
-RUN pip install 'pytest>=3' coverage coveralls
+RUN pip install 'pytest>=3' coverage
 
 # This will invalidate the cache if something changes in python_moztelemetry.
 COPY . /python_moztelemetry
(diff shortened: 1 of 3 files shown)


fbertsch:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1414076

runtests.sh
```diff
@@ -17,8 +17,3 @@ if [ $# -gt 0 ]; then
 else
     coverage run --source=moztelemetry setup.py test
```
robhudson: Should we not run `coverage` until we have a replacement?

Dockerfile
```diff
@@ -5,7 +5,8 @@ ENV SPARK_VERSION=2.0.2
 # install gcc
 RUN apt-get update --fix-missing && \
     apt-get install -y \
-    g++ libpython-dev libsnappy-dev
+    g++ libpython-dev libsnappy-dev \
+    build-essential libssl-dev libffi-dev
```
robhudson: Why the extra packages?

fbertsch: These are dependencies for `cryptography`, see the second commit.

fbertsch:
@robhudson RFAL

PR #187, opened 2017-11-09
Bug 1414582 - Print warning for "sample"

relud:
Fixes Bug 1414582 Print a warning when using the "sample" functionality of Dataset

r? @mreid-moz

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -253,6 +253,12 @@ class Dataset:
         if summaries and limit is None and sample != 1:
             if sample < 0 or sample > 1:
                 raise ValueError('sample must be between 0 and 1')
+            print(
+                "WARNING: THIS IS NOT A REPRESENTATIVE SAMPLE.\n"
+                "This 'sampling' is based on s3 files and is highly\n"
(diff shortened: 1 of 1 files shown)


mreid-moz:
Looks good to me, r? @sunahsuh for the copy of the warning.

moztelemetry/dataset.py
```diff
@@ -253,6 +253,11 @@ def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None
         if summaries and limit is None and sample != 1:
             if sample < 0 or sample > 1:
                 raise ValueError('sample must be between 0 and 1')
+            print(
+                "WARNING: sampling based on the set of files in s3. This\n"
```
sunahsuh: I think the first line should very explicitly be: this is not a representative sample, and we should be explicit about the implications e.g.

 "WARNING: THIS IS NOT A REPRESENTATIVE SAMPLE. This "sampling" is based on s3 files and is highly susceptible to skew. Use only for quicker performance while prototyping."

relud:
okay, updated the warning

sunahsuh:
👍 :shipit:

PR #189, opened 2018-01-03
Add test for extracted fields that are nested

acmiyaguchi:
There doesn't seem to be a test for nested fields names like "foo.bar.baz". This tests the recursive behavior of flattened fields that are nested several levels.


acmiyaguchi:
:+1:

PR #191, opened 2018-02-01
Bug 1376905 - Run flake8 when testing locally

robhudson:


acmiyaguchi:
How do you run this?

robhudson:
It kicks in with the `./runtests.sh` command. Or `python setup.py test` in the container.

acmiyaguchi:
The broken flake8 tests need to be fixed before this can be merged.

robhudson:
Not having written the code fixing the bare `except`s may need to be a team effort. It appears that the flake8 as it was set up before wasn't catching these. These may be a recent addition to the flake8 checks.

I also updated the travis.yml file to remove the test matrix since flake8 is now run as part of the tests themselves instead of needing a separate run.

PR #194, opened 2018-02-06
Use parse_scalars.py instead of custom code

Dexterp37:
This additionally removes the `REQUIRED_FIELDS` and `OPTIONAL_FIELDS` dictionaries: these checks would
be performed by the `parse_scalars.py` library with `strict_type_checks=True`. However, for server side computation, we're usually disabling this to be backward compatible with older registry formats.


Dexterp37:
There you go :)

bin/update_parser_scripts
```diff
@@ -0,0 +1,35 @@
...
+
+    print("{} downloaded.".format(file_name))
+
+    with open(os.path.join(target_dir, file_name), "w") as fh:
+        fh.write(response.content)
+
+    print("{} successfully updated.".format(file_name))
+
+# Update the required files.
+FILES_TO_UPDATE = [
```
fbertsch: nit: move this to the top of the file

PR #208, opened 2018-05-02
Add support for python3

relud:
and make python3 default in docker

### DIFF
--- a/moztelemetry/heka/message_parser.py
+++ b/moztelemetry/heka/message_parser.py
@@ -5,12 +5,12 @@ import gzip
 import ssl
 import struct
 
-import boto
+import boto.s3.key
 import snappy
 import ujson as json
 import json as standard_json
 import zlib
-from cStringIO import StringIO
+from io import BytesIO
 from google.protobuf.message import DecodeError
 
 from .message_pb2 import Message, Header
@@ -42,17 +42,18 @@ def _parse_heka_record(record):
             # messages is an unprocessed form of the data, usually the original
             # gzipped payload from the client.
             #
-            # We attempt to decompress it, and if that fails,
-            # attempt to decode it as a UTF-8 string.
+            # We decompress it if we can, then try to decode it as a UTF-8 string.
             elif field.name == 'content':
                 try:
-                    string = zlib.decompress(field.value_bytes[0], 16+zlib.MAX_WBITS)
-                except zlib.error:
+                    string = field.value_bytes[0]
                     try:
-                        string = field.value_bytes[0].decode('utf-8')
-                    except UnicodeDecodeError:
-                        # There is no associated payload
-                        break
+                        string = zlib.decompress(string, 16+zlib.MAX_WBITS)
+                    except zlib.error:
+                        pass  # not compressed
+                    string = string.decode('utf-8')
+                except UnicodeDecodeError:
+                    # There is no associated payload
+                    break
                 payload = {"content": string}
                 break
 
@@ -123,7 +124,7 @@ _record_separator = 0x1e
 class BacktrackableFile:
     def __init__(self, stream):
         self._stream = stream
-        self._buffer = StringIO()
+        self._buffer = BytesIO()
 
     def read(self, size):
         buffer_data = self._buffer.read(size)
@@ -151,7 +152,7 @@ class BacktrackableFile:
         buf = self._buffer.getvalue()
         index = buf.find(chr(_record_separator), 1)
 
-        self._buffer = StringIO()
+        self._buffer = BytesIO()
         if index >= 0:
             self._buffer.write(buf[index:])
             self._buffer.seek(0)
@@ -170,7 +171,7 @@ def read_until_next(fin, separator=_record_separator):
     bytes_skipped = 0
     while True:
         c = fin.read(1)
-        if c == '':
+        if len(c) == 0:
             return bytes_skipped, True
         elif ord(c) != separator:
             bytes_skipped += 1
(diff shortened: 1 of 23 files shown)


moztelemetry/heka/message_parser.py
```diff
@@ -205,9 +205,9 @@ def read_one_record(input_stream, raw=False, verbose=False, strict=False, try_sn
     header.ParseFromString(header_raw)
     unit_separator = input_stream.read(1)
     total_bytes += 1
-    if ord(unit_separator[0]) != 0x1f:
+    if ord(unit_separator) != 0x1f:
```
fbertsch: Did this fail with the `[0]` left in? It seems to me it should behave the same way with or without `[0]` and with `StringIO` or `BytesIO`, but I wanted to check and see.

moztelemetry/histogram.py
```diff
@@ -8,15 +8,25 @@
...
+from . import parse_histograms
 from expiringdict import ExpiringDict
 
+# python 2 and 3 compatiblity
+try:
+    from functools32 import lru_cache
+except ImportError:
+    from functools import lru_cache
+
+try:
```
fbertsch: This should use `from past.builtins import long` (and keep code as-is) or `from builtins import int` (and change the `long` to `int`)

moztelemetry/parse_histograms.py
```diff
@@ -9,11 +9,17 @@
...
-import shared_telemetry_utils as utils
 
-from shared_telemetry_utils import ParserError
+from . import shared_telemetry_utils as utils
+from .shared_telemetry_utils import ParserError
 from collections import OrderedDict
 
+# python 2 and 3 compatiblity
+try:
+    basestring
```
fbertsch: Can use `from past.builtins import basestring`

moztelemetry/spark.py
```diff
@@ -35,6 +42,11 @@ class PingCursor(dict):
     for more details.
     """
 
+    try:
+        viewkeys = dict.viewkeys
```
fbertsch: can do `from future.utils import viewkeys`.

moztelemetry/heka/message_parser.py
relud: in python 3 if you index bytes, you get a number instead of bytes, where in python 2.7 you get a string. but it's already length 1 because of `input_stream.read(1)`, so the `[0]` was unnecessary.

fbertsch: Interesting, thanks!

wlach:
Should we merge this?

moztelemetry/parse_histograms.py
```diff
@@ -616,7 +617,7 @@ def from_files(filenames, strict_type_checks=True):
         if not isinstance(histograms, OrderedDict):
             raise ParserError("Histogram parser did not provide an OrderedDict.")
 
-        for (name, definition) in histograms.iteritems():
+        for (name, definition) in histograms.items():
```
jezdez: Remove the parentheses here.

moztelemetry/parse_histograms.py
```diff
@@ -642,5 +643,5 @@ def from_files(filenames, strict_type_checks=True):
                   'histogram-whitelists.json:\n%s'
             raise ParserError(msg % (', '.join(sorted(orphaned))))
 
-    for (name, definition) in all_histograms.iteritems():
+    for (name, definition) in all_histograms.items():
```
jezdez: Parentheses as well.

moztelemetry/store.py
```diff
@@ -99,7 +99,9 @@ def get_key(self, key):
             value = self.store[key]
         except KeyError:
             raise Exception('Error retrieving key "{}" from S3'.format(key))
-        output = StringIO.StringIO()
+        output = BytesIO()
+        if type(value) is not bytes:
+            value = value.encode('utf-8')
```
jezdez: You can use [`six.ensure_str`](https://github.com/benjaminp/six/blob/68112f3193c7d4bef5ad86ed1b6ed528edd9093d/six.py#L870-L887) here.

moztelemetry/histogram.py
```diff
@@ -8,14 +8,20 @@
...
-from functools32 import lru_cache
+from . import parse_histograms
 from expiringdict import ExpiringDict
+from past.builtins import long
+
+# python 2 and 3 compatiblity
+try:
+    from functools32 import lru_cache
+except ImportError:
+    from functools import lru_cache
```
jezdez: Nice!

moztelemetry/spark.py
```diff
@@ -148,15 +151,15 @@ def get_pings_properties(pings, paths, only_median=False, with_processes=False,
     keyed by the original paths (if 'paths' is a list) or the custom identifier keys
     (if 'paths' is a dict).
     """
-    if isinstance(pings.first(), str):
-        pings = pings.map(lambda p: json.loads(p))
+    if isinstance(pings.first(), bytes):
```
jezdez: Best would be to use `six.binary_type` for this check.

runtests.sh
```diff
@@ -3,13 +3,19 @@
...
 if [ ! -f /.dockerenv ]; then
-    docker run -t -i -v $PWD:/python_moztelemetry moztelemetry_docker ./runtests.sh "$@"
+    docker run -t -i -v $PWD:/python_moztelemetry -e PYTHON_VERSION="$PYTHON_VERSION" moztelemetry_docker ./runtests.sh "$@"
     exit $?
 fi
 
+if [ "$PYTHON_VERSION" == '2.7' ]; then
+    export PATH="$PYTHON27_PATH"
+fi
+
```
jezdez: I'm wondering whether we should allow testing against different Python 3 versions as well instead of hardcoding 2.7 and 3.5.

My experience with Conda is limited so I'm not sure if we're depending on the system Python version of the openjdk Docker image (which I think is Debian stretch). Can we try to add 3.6 to the list of Python's and see if it creates a conda env successfully with that?

moztelemetry/store.py
relud: in this case i'm trying to ensure bytes, in python 3

jezdez: Whoops, that's what I meant as well, tripped over the naming thing heh

jezdez:
@relud 😀

PR #209, opened 2018-05-02
 Bug 1447851 - Add content decoding for landfill

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1447851)

Option b seemed like a good choice. Landfill messages will be parsed directly into a string, which can be later decoded using `json.loads` if needed.

### DIFF
--- a/moztelemetry/heka/message_parser.py
+++ b/moztelemetry/heka/message_parser.py
@@ -9,6 +9,7 @@ import boto
 import snappy
 import ujson as json
 import json as standard_json
+import zlib
 from cStringIO import StringIO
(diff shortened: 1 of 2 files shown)


whd:
This is indeed the "weird case" I mentioned in my previous comment; my intent was to suggest that we [do what we already do when decoding normal fields](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/heka/message_parser.py#L60-L64) (effectively, drop the field).

acmiyaguchi:
Merging this in, I had been waiting for Python 3 support, but it may be better to rebase on this new functionality.

PR #216, opened 2018-07-18
Bug 1451103 - add method for partition strategy

erindcole:
Wondering if I could get some feedback on this method. It implements a priority queue and groups the data with a list of lists.

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -42,29 +42,29 @@ def _group_by_size_greedy(obj_list, tot_groups):
     return groups
 
 def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
-  sub_list = []
-  groups = [[]]
-  pq = []
-  tmp_object = none
-  
-  if tot_groups <= 1:
-    group_by_size_greedy(obj_list, tot_groups)
-  
-  pq = heapq.heapify(obj_list)
-  
-  while pq: 
-    if pq[0].size >= threshold:
-      tmp_object = heapq.heappop(pq)
-      groups.append(tmp_object)
-    else:
-      if sub_list.size >= threshold:
-        groups.append(sub_list)
-        sub_list = []
-      else:
-        tmp_object = heapq.heappop(pq)
-        sub_list.append(tmp_object)
+    sub_list = []
+    groups = [[]]
+    pq = []
+    tmp_object = none
 
-  return groups
+    if tot_groups <= 1:
+        group_by_size_greedy(obj_list, tot_groups)
+
+    pq = heapq.heapify(obj_list)
+
+    while pq: 
+        if pq[0].size >= threshold:
+            tmp_object = heapq.heappop(pq)
+            groups.append(tmp_object)
+    else:
+        if sub_list.size >= threshold:
+            groups.append(sub_list)
+            sub_list = []
+        else:
+            tmp_object = heapq.heappop(pq)
+            sub_list.append(tmp_object)
+
+    return groups
 
 def _pickle_method(m):
     """Make instance methods pickable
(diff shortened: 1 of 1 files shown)


moztelemetry/dataset.py
```diff
@@ -40,6 +41,30 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
     return groups
 
+def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
+    sub_list = []
+    groups = [[]]
+    pq = []
+    tmp_object = none
+
+    if tot_groups <= 1:
+        group_by_size_greedy(obj_list, tot_groups)
```
acmiyaguchi: typo: There should be an underscore prefixing this function

moztelemetry/dataset.py
```diff
@@ -40,6 +41,30 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+
+    pq = heapq.heapify(obj_list)
```
acmiyaguchi: Items in the priority queue should be tuples of `(priority, object)` as per the [second example](https://docs.python.org/2/library/heapq.html#basic-examples) in the heapq docs.

moztelemetry/dataset.py
```diff
@@ -72,7 +97,7 @@ class Dataset:
         schema = ['submissionDate', 'docType', 'platform']
 
         records = Dataset(bucket, schema) \\
-            .select(
+                .select(
```
acmiyaguchi: I think these are accidental whitespacing changes from a formatter. I would use `git add -p` to manually inspect the changes that you want to make.

moztelemetry/dataset.py
```diff
@@ -40,6 +41,30 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+    if tot_groups <= 1:
+        group_by_size_greedy(obj_list, tot_groups)
+
+    pq = heapq.heapify(obj_list)
+
+    while pq: 
+        if pq[0].size >= threshold:
+            tmp_object = heapq.heappop(pq)
+            groups.append(tmp_object)
+    else:
```
acmiyaguchi: Typo: proper indentation

moztelemetry/dataset.py
```diff
@@ -40,6 +41,30 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+        if sub_list.size >= threshold:
```
acmiyaguchi: The `sub_list` doesn't have a `.size` method.

moztelemetry/dataset.py
erindcole: I use the heapify() method to make the heap, found [here](https://docs.python.org/2/library/heapq.html#heapq.heapify). It takes a list and sorts it in place. I believe the default behavior is to prioritize by smallest.

moztelemetry/dataset.py
erindcole: Okay I will, thanks.

moztelemetry/dataset.py
acmiyaguchi: I think heapify needs a comparison operator like `.__cmp__` or `.__lt__` to be defined. Dictionaries are like hash-tables, and don't have a natural order like integers or strings do.

moztelemetry/dataset.py
erindcole: Thanks for catching this. I meant to look that up for proper syntax.

erindcole:
Thanks for the comments and feedback. I will fix the smaller errors and will test the method better. How do you suggest I run the unit tests? It seems I almost need to have this method included in the repo first in order to fully test its performance against my Benchmark tests.

acmiyaguchi:
You can add your tests around here.

https://github.com/mozilla/python_moztelemetry/blob/4148c5f64d5ee0e669815922579ac3e87f33bc8e/tests/test_dataset.py#L221-L238

The tests for the new method will probably look similar to this.

PR #217, opened 2018-07-24
Bug 1463877 - partition strategy

erindcole:
Could I get some feedback on these changes? This method groups files into equally sized groups, as an alternative to the greedy strategy.

The method passes the unit tests. Next step is to test it against the benchmark tests after it is included.

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -43,6 +43,7 @@ def _group_by_size_greedy(obj_list, tot_groups):
 
 
 def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
+<<<<<<< HEAD
     """Partition a list of objects evenly and by file size
 
     Files are placed according to largest file in the smallest bucket. If the
@@ -55,6 +56,10 @@ def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
     """
     sorted_obj_list = sorted([(obj['size'], obj) for obj in obj_list], reverse=True)
     groups = [(random.random(), []) for _ in range(tot_groups)]
+=======
+    sorted_obj_list = sorted([(obj['size'], obj) for obj in obj_list], reverse=True)
+    groups = [(0, []) for _ in range(tot_groups)]
+>>>>>>> dfc73a9... improve style
 
     if tot_groups <= 1:
         groups = _group_by_size_greedy(obj_list, tot_groups)
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -259,7 +259,7 @@ class Dataset:
         keys = sc.parallelize(scanned).flatMap(self.store.list_keys)
         return keys.take(limit) if limit else keys.collect()
 
-    def records(self, sc, group_by, limit=None, sample=1, seed=42, decode=None, summaries=None):
+    def records(self, sc, group_by='greedy', limit=None, sample=1, seed=42, decode=None, summaries=None):
         """Retrieve the elements of a Dataset
 
         :param sc: a SparkContext object
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -43,7 +43,6 @@ def _group_by_size_greedy(obj_list, tot_groups):
 
 
 def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
-<<<<<<< HEAD
     """Partition a list of objects evenly and by file size
 
     Files are placed according to largest file in the smallest bucket. If the
@@ -56,10 +55,6 @@ def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
     """
     sorted_obj_list = sorted([(obj['size'], obj) for obj in obj_list], reverse=True)
     groups = [(random.random(), []) for _ in range(tot_groups)]
-=======
-    sorted_obj_list = sorted([(obj['size'], obj) for obj in obj_list], reverse=True)
-    groups = [(0, []) for _ in range(tot_groups)]
->>>>>>> dfc73a9... improve style
 
     if tot_groups <= 1:
         groups = _group_by_size_greedy(obj_list, tot_groups)
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -50,7 +50,7 @@ def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
     by itself.
     :param obj_list: a list of dict-like objects with a 'size' property
     :param tot_groups: number of partitions to split the data
-    :threshold: the maximum size of each bucket
+    :param threshold: the maximum size of each bucket
     :return: a list of lists, one for each partition
     """
     sorted_obj_list = sorted([(obj['size'], obj) for obj in obj_list], reverse=True)
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -43,35 +43,28 @@ def _group_by_size_greedy(obj_list, tot_groups):
 
 
 def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
-    sub_list = []
-    groups = []
-    priority = []
-    tmp_object = None
-    sub_list_size = 0
+    groups = [[]]
+    sorted_obj_list = []
+    best_partition = None
 
     if tot_groups <= 1:
         groups = _group_by_size_greedy(obj_list, tot_groups)
         return groups
-
-    for obj in obj_list:
-        for key in obj:
-            priority.append((obj[key], obj))
-
-    heapq.heapify(priority)
-    while priority:
-        tmp_object = heapq.heappop(priority)
-
-        if tmp_object[0] >= threshold:
-            groups.append(tmp_object)
-
-        sub_list.append(tmp_object)
-        sub_list_size += tmp_object[0]
-        if sub_list_size >= threshold:
-            groups.append(sub_list)
-            sub_list = []
-            sub_list_size = 0
-    if sub_list:
-        groups.append(sub_list)
+    sorted_obj_list = [(obj['size'], obj) for obj in obj_list]
+    sorted_obj_list.sort(reverse=True)
+    groups = [(0, []) for _ in range(tot_groups)]
+    heapq.heapify(groups)
+    for obj in sorted_obj_list:
+        if obj[0] > threshold:
+            heapq.heappush(groups, (obj[0], [obj[1]]))
+        else:
+            best_partition = heapq.heappop(groups)
+            (size, files) = best_partition
+            size += obj[0]
+            files.append(obj[1])
+            best_partition = (size, files)
+            heapq.heappush(groups, best_partition)
+    groups = [group[1] for group in groups]
     return groups
 
 
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -3,6 +3,7 @@
 # file, you can obtain one at http://mozilla.org/MPL/2.0/.
 from __future__ import division, print_function
 import functools
+import heapq
 import json
 import random
 import re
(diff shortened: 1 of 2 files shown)


moztelemetry/dataset.py
```diff
@@ -41,6 +42,39 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+    groups = []
+    priority = []
+    tmp_object = None
+    sub_list_size = 0
+
+    if tot_groups <= 1:
+        groups = _group_by_size_greedy(obj_list, tot_groups)
+        return groups
+
+    for obj in obj_list:
```
acmiyaguchi: List comprehensions are a nice compact way of doing this for-loop:

```python
priority = [(obj[key], obj) for obj in obj_list]
```

moztelemetry/dataset.py
```diff
@@ -41,6 +42,39 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+        for key in obj:
```
acmiyaguchi: The only key we're interested in using is the `size` key.

moztelemetry/dataset.py
```diff
@@ -41,6 +42,39 @@ def _group_by_size_greedy(obj_list, tot_groups):
...
+    while priority:
+        tmp_object = heapq.heappop(priority)
+
+        if tmp_object[0] >= threshold:
+            groups.append(tmp_object)
+
+        sub_list.append(tmp_object)
+        sub_list_size += tmp_object[0]
+        if sub_list_size >= threshold:
+            groups.append(sub_list)
```
acmiyaguchi: I think the sub-lists should also be in a heap so that the next largest item gets added to the smallest list. The group should be a heap of length `tot_groups` e.g `[(0, []) for _ in range(tot_groups)]`.

Here's an example of where this approach won't pack optimally. Each row represents the successive sizes of the partitions. 

total groups = 3
threshold = 5
file sizes = 2, 2, 2

part 1| part 2| part 3
--|--|--
 0 | 0 | 0 
2 | 0 | 0
4 | 0 | 0
4 | 2 | 0

Instead, all of the partitions should be as evenly sized as possible. One possible way would be to add the next largest item to the emptiest partition:

part 1| part 2| part 3
--|--|--
 0 | 0 | 0 
2 | 0 | 0
2 | 2 | 0
2 | 2 | 2

tests/test_dataset.py
```diff
@@ -238,6 +239,27 @@ def test_group_by_size_greedy():
...
+    obj_list2 = [{'size': 70}, {'size': 70}, {'size': 70}, {'size': 70}]
+    obj_list3 = [{'size': 4}, {'size': 1}, {'size': 3}, {'size': 2}]
+
+    groups = _group_by_equal_size(obj_list1, 1)
+    assert groups == [
+        [{'size': 4}, {'size': 3}, {'size': 2}, {'size': 1}]
+    ]
+    groups = _group_by_equal_size(obj_list2, 2, 100)
+    assert groups == [
+        [(70, {'size': 70}), (70, {'size': 70})],
```
acmiyaguchi: Why does the first assertion take a list of dictionaries while this one takes a list of tuples? The return types seem to conflict.

erindcole: I understood the desired output for `_group_by_equal_size()` to be a list of tuples. I can change this to a list of dictionaries. Do you prefer the desired output to be a list of dictionaries so it is the same type as the greedy approach? The first assertion is meant to test the behavior in the case where the greedy approach is called within `_group_by_equal_size()`.

moztelemetry/dataset.py
erindcole: Okay, this makes sense. If this is the desired behavior, it does make more sense to use a second heap instead of a list. However, it seems I would then need to check the size of each group on every iteration in order to keep the groups equally sized. Is this true? Wanted to make sure I am on the right track with the desired implementation.

tests/test_dataset.py
acmiyaguchi: The return should be the same as `_group_by_equal_size` since it should be a drop in replacement for the use in [Dataset.records()](https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/dataset.py#L278-L279). This might crash during runtime since its a different type.

The list of tuples is only needed for the sorting in the heapq methods.

moztelemetry/dataset.py
acmiyaguchi: Correct, there the size of the group needs to be checked on every iteration. If the size of the next item is added to the running group size, the heap can take care of the ordering though.

moztelemetry/dataset.py
erindcole: The problem is obj_list is a list of dictionaries. It seems list comprehensions do not work in this case. In order to access the key of each dictionary, I thought I needed to index into the list first.

acmiyaguchi: It's doable if there's only one key being extracted. Here's a toy example.

```python
>>> test = [{'a': x, 'b': x**2} for x in range(3)]

>>> print(test)
[{'a': 0, 'b': 0}, {'a': 1, 'b': 1}, {'a': 2, 'b': 4}]

>>> [(x['b'], x) for x in test]
[(0, {'a': 0, 'b': 0}), (1, {'a': 1, 'b': 1}), (4, {'a': 2, 'b': 4})]
```

erindcole: This makes sense. It also works because you are able to specify a key that is known in the dictionary in the toy example. I will research more into list comprehensions with lists of dictionaries. However, it may be tricky to adapt the needed modifications to make a list comprehension work in this specific case.

acmiyaguchi: Are there other keys aside from the `size` key being used in the file summaries?

erindcole: It does work. Thanks for the clarification.

tests/test_dataset.py
```diff
@@ -238,6 +239,41 @@ def test_group_by_size_greedy():
...
+    groups = _group_by_equal_size(obj_list3, 3, 5)
+    assert groups == [
+        ['size': 2}, {'size': 1}],
+        [{'size': 4}],
+        [{'size': 3}]
+    ]
+    groups = _group_by_equal_size(obj_list4, 3, 5)
+    assert groups == [
+        [{'size': 2}],
+            [{'size': 2}],
```
acmiyaguchi: Nit: indentation is off

moztelemetry/dataset.py
```diff
@@ -41,6 +42,32 @@ def _group_by_size_greedy(obj_list, tot_groups):
     return groups
 
 
+def _group_by_equal_size(obj_list, tot_groups, threshold=pow(2, 32)):
```
acmiyaguchi: Add a docstring like `_group_by_size_greedy`.

moztelemetry/dataset.py
```diff
@@ -231,7 +258,8 @@ def summaries(self, sc, limit=None):
         keys = sc.parallelize(scanned).flatMap(self.store.list_keys)
         return keys.take(limit) if limit else keys.collect()
 
-    def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None):
+    def records(self, sc, limit=None, sample=1, seed=42, decode=None,
+            summaries=None, group_by='greedy', 'equal_size'):
```
acmiyaguchi: There should only be value: `group_by='greedy'`. Please add to the doc-string to reflect the updated parameters.

moztelemetry/dataset.py
```diff
@@ -274,7 +302,12 @@ def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None
         total_size_mb = total_size / float(1 << 20)
         print("fetching %.5fMB in %s files..." % (total_size_mb, len(summaries)))
 
-        groups = _group_by_size_greedy(summaries, 10 * sc.defaultParallelism)
+        if group_by == 'equal_size':
+            groups = _group_by_equal_size(summaries, 10*sc.defaultParallelism)
+        elif group_by == 'greedy':
+            groups = _group_by_size_greedy(summaries, 10 * sc.defaultParallelism)
+        else:
+            groups = _grou_by_size_greedy(summaries, 10*sc.defaultParallelism)
```
acmiyaguchi: typo: _grou_by_size_greedy

moztelemetry/dataset.py
```diff
@@ -231,7 +258,7 @@ def summaries(self, sc, limit=None):
         keys = sc.parallelize(scanned).flatMap(self.store.list_keys)
         return keys.take(limit) if limit else keys.collect()
 
-    def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None):
+    def records(self, sc, group_by, limit=None, sample=1, seed=42, decode=None, summaries=None):
```
acmiyaguchi: `group_by` should have a single default value: `group_by='greedy'`. Can you add the new parameter to the docstring?

erindcole:
Wanted to clarify. Your comments are largely asking for style changes. Am I correct about this? Wanted to understand the improvements made to the implementation by making these changes. It seems these changes are meant to enforce consistency for the code base.

acmiyaguchi:
Yes, a majority of the changes are for documentation and concision of code. There is an important change that needs to be made to the `.records()` method that would otherwise break existing code.

moztelemetry/dataset.py
erindcole: Could you clarify what you mean by this? Is there a particular format I should follow that is generally used, or should I mostly copy the style of the _group_by_size_greedy format as closely as possible?

acmiyaguchi: Docstrings are the pythonic way of docmenting code. See [PEP 257](https://www.python.org/dev/peps/pep-0257/) for some background. Concretely, we have [autogenerated doc pages](https://python-moztelemetry.readthedocs.io/en/latest/) for this repository that are created from docstrings.

We use [sphinx](http://www.sphinx-doc.org/en/master/) to package the documentation, you can see the `/docs` folder for how the build process works. If you follow the general style of the comments that exist in the repo, it should probably work fine. Here's [a website](https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html#auto-documented-class-method-1) that has some examples on some of the syntax like `:param <argument name>:`. I've also provided an example with the elements of a good docstring. 

```python
def function(arg1, arg2):
    """Here's a short description.

    Here's a more detailed description. It can have multiple sentences. It
    may describe context and why you may use this function.

    :param arg1: the type and description
    :param arg2: the type and description
    :return: a type and description
    """
    pass
```

tests/test_dataset.py
```diff
@@ -238,6 +239,41 @@ def test_group_by_size_greedy():
...
+        [{'size': 4}, {'size': 3}, {'size': 2}, {'size': 1}]
+    ]
+    groups = _group_by_equal_size(obj_list2, 2, 100)
+    assert groups == [
+        [{'size': 70}, {'size': 70}],
+        [{'size': 70}, {'size': 70}]
+    ]
+    groups = _group_by_equal_size(obj_list3, 3, 5)
+    assert groups == [
+        ['size': 2}, {'size': 1}],
```
acmiyaguchi: This is missing an opening curly brace.

tests/test_dataset.py
```diff
@@ -238,6 +239,41 @@ def test_group_by_size_greedy():
...
+        [{'size': 2}],
+            [{'size': 2}],
+                [{'size': 2}]
+    ]
+    groups = _group_by_equal_size(obj_list5, 2, 100)
+    assert groups == [
+        [{'size': 70}],
+            [{'size':150}],
+                [{'size': 70}, {'size': 70}]
+
```
acmiyaguchi: This is missing a closing bracket.

moztelemetry/dataset.py
```diff
@@ -275,7 +306,12 @@ def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None
         total_size_mb = total_size / float(1 << 20)
         print("fetching %.5fMB in %s files..." % (total_size_mb, len(summaries)))
 
-        groups = _group_by_size_greedy(summaries, 10 * sc.defaultParallelism)
+        if group_by == 'equal_size':
+            groups = _group_by_equal_size(summaries, 10*sc.defaultParallelism)
+        elif group_by == 'greedy':
+            groups = _group_by_size_greedy(summaries, 10 * sc.defaultParallelism)
+        else:
+            groups = _group_by_size_greedy(summaries, 10*sc.defaultParallelism)
```
acmiyaguchi: Can you print something along the lines of "this group_by method is invalid, defaulting to greedy"? Alternatively, raising an exception on an invalid option. There isn't any feedback on whether it works in this case.

moztelemetry/dataset.py
```diff
@@ -232,7 +263,7 @@ def summaries(self, sc, limit=None):
         keys = sc.parallelize(scanned).flatMap(self.store.list_keys)
         return keys.take(limit) if limit else keys.collect()
 
-    def records(self, sc, limit=None, sample=1, seed=42, decode=None, summaries=None):
+    def records(self, sc, group_by='greedy', limit=None, sample=1, seed=42, decode=None, summaries=None):
```
acmiyaguchi: Can you also add `group_by` to the docstring?

moztelemetry/dataset.py
erindcole: To clarify, you would prefer an exception be raised in the `else` statement rather than a call to the greedy strategy. Am I correct with my understanding?

PR #231, opened 2018-09-13
Use a named docker volume for tox cache

jklukas:
This should avoid the problem of files being created with root ownership on Linux hosts.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,6 @@
 
 .*
 *.egg*
+**/__pycache__/
+**.pyc
+**.pyo
--- a/bin/clean
+++ b/bin/clean
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Remove various cached files that are intended to improve performance
+# but can cause errors as the environment evolves.
+
+# Move to the root of the git repository.
+cd $(git rev-parse --show-toplevel)
+
+# Clean all python cache files.
+find moztelemetry/ tests/ | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
+
+# Remove any local tox environments.
+rm -rf .*tox/
+
+# Delete the docker volume we use for caching tox environment.
+docker volume rm moztelemetry-tox
--- a/bin/test
+++ b/bin/test
@@ -8,22 +8,23 @@ cd $(git rev-parse --show-toplevel)
 
 # Set default variables if not already present in the environment.
 : ${PYTHON_VERSION:=3.6}
-: ${TOXDIR:=.dockertox}  # Prevents the circleci cli from picking up the env
 : ${TOXENV:="py${PYTHON_VERSION/./}"}
 
 # Set some derived variables.
 TOXENV="py${PYTHON_VERSION/./}"
 
-# Clean all python cache files, since they may confuse the docker environment.
-find moztelemetry/ tests/ | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
+# Create a named volume to hold cached tox environment between runs;
+# this can cause errors with running tests after modifying setup.py, etc.,
+# so be prepared to run the bin/clean script.
+docker volume create moztelemetry-tox
 
 # If nothing has changed, this build step will hit cache and run quickly.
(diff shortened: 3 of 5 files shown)


bin/test
```diff
@@ -8,22 +8,23 @@ cd $(git rev-parse --show-toplevel)
...
 
 # If nothing has changed, this build step will hit cache and run quickly.
 docker build -t moztelemetry:"$TOXENV" --build-arg PYTHON_VERSION="$PYTHON_VERSION" .
 
 # We mount the current directory to the container and run tox inside it.
 docker run -it --rm \
-       --volume "$(pwd)/$TOXDIR":/python_moztelemetry/"$TOXDIR" \
-       --volume "$(pwd)"/.git:/python_moztelemetry/.git \
+       --mount source=moztelemetry-tox,target=/python_moztelemetry/.tox \
+       --mount type=bind,source="$(pwd)"/.git,target=/python_moztelemetry/.git,readonly \
```
jklukas: Apparently, the `--mount` option is now preferred over `--volume`. It uses a more verbose syntax, but is more flexible and explicit about what's going on.

In particular, it makes it more clear that one of these is a bind mount of a directory on the host and the other one isn't.

acmiyaguchi:
I can get this into a strange state by running tox locally before running the test script. I can run the clean script before the test script to get rid of the errors. Other than that, I'm glad the permission errors aren't around anymore. It looks like the docker container effectively gets it's own mounted drive that's isolated from the host file system.

```bash
# `-k test_dates` limits the number of tests run
$ tox -e py36 -- -k test_dates
$ ./bin/tests -k test_dates

...
Traceback (most recent call last):
  File "/python_moztelemetry/.tox/py36/lib/python3.6/site-packages/_pytest/config/__init__.py", line 381, in _getconftestmodules
    return self._path2confmods[path]
KeyError: local('/python_moztelemetry/tests')

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/python_moztelemetry/.tox/py36/lib/python3.6/site-packages/_pytest/config/__init__.py", line 412, in _importconftest
    return self._conftestpath2mod[conftestpath]
KeyError: local('/python_moztelemetry/tests/conftest.py')

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/python_moztelemetry/.tox/py36/lib/python3.6/site-packages/_pytest/config/__init__.py", line 418, in _importconftest
    mod = conftestpath.pyimport()
  File "/python_moztelemetry/.tox/py36/lib/python3.6/site-packages/py/_path/local.py", line 686, in pyimport
    raise self.ImportMismatchError(modname, modfile, self)
py._path.local.LocalPath.ImportMismatchError: ('tests.conftest', '/home/anthony/Work/python_moztelemetry/tests/conftest.py', local('/python_moztelemetry/tests/conftest.py'))
ERROR: could not load /python_moztelemetry/tests/conftest.py

ERROR: InvocationError for command '/python_moztelemetry/.tox/py36/bin/pytest -k test_dates' (exited with code 4)
```

bin/clean
```diff
@@ -0,0 +1,19 @@
...
+cd $(git rev-parse --show-toplevel)
+
+# Clean all python cache files.
+find moztelemetry/ tests/ | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
+
+# Remove any local tox environments.
+rm -r .*tox/
+
+# Delete the docker volume we use for caching tox environment.
+docker volume create moztelemetry-tox
```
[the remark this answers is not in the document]
jklukas: Yes, thank you. Will fix.

.dockerignore
```diff
@@ -3,3 +3,6 @@
 
 .*
 *.egg*
+**/__pycache__/
+**.pyc
+**.pyo
```
jklukas: I was able to reproduce @acmiyaguchi's issue on my machine and then solve the problem by using `**` to ignore these caches in arbitrary subdirectories.

jklukas:
Squashed in a few fixups that should address the outstanding issues in this PR, and an additional commit that pins boto3 version to get tests passing.

Merged to avoid drift, but I'd be happy to issue a follow-up PR if you have any concerns about the post-review changes.

PR #235, opened 2018-10-30
Bug 1341566 Regenerate protobuf classes

jklukas:
Also adds documentation on how to regenerate the classes in the future.


fbertsch:
@jklukas have you tested this on some data from the data lake?

jklukas:
> have you tested this on some data from the data lake?

I was relying on the heka tests to verify this update is working fine, but I can test on real data from the lake. I'm spinning up an ATMO cluster and I'll try out the telemetery hello world with these updates in place.

fbertsch:
Be wary that the worker nodes will not have an updated version of this unless you ssh into them and updated the lib separately. I would test on a single node, reading a few small files should be enough.

jklukas:
Good call on testing. I ssh'ed this updated file into the python environment and received an error in the Hello World notebook about not recognizing "serialized options". I'm pretty sure this just means we need an updated version of the `protobuf` library. Testing that out now, and we'll need to issue a follow-up PR to spark bootstrap to update moztelemetry and protobuf.

jklukas:
Hello World is working with the code of this PR and updated pandas and protobuf. Will add in a lower bound on protobuf version.

jklukas:
Ready for another look, @fbertsch.

PR #237, opened 2018-11-01
Test docs and deploy to gh-pages branch

jklukas:
We should definitely be testing docs as part of CI to make sure they build,
which is addressed here. But this change also explores what it could look
like to publish docs to Github Pages rather than ReadTheDocs, so that we can
avoid the additional developer friction of understanding that service and
maintaining user permissions there.

The gh-pages docs are live at:
http://mozilla.github.io/python_moztelemetry/

A few features I notice missing that ReadTheDocs provides:

- hosting multiple versions of the docs, though we don't look to be using this
- download links for PDF, HTML, and Epub
- "Edit on GitHub" links; the gh-pages rendered version links to the content
  on the gh-pages branch rather than on master

The above features are indeed nice to have. RTD is also a fairly python-specific tool, so if we do value hosting API docs for our projects, the technique here is a bit more transferable.

This PR is mostly intended to provoke discussion. I'm totally fine if we decide to close this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,15 +20,15 @@
 test_settings: &test_settings
   steps:
     - checkout
-    - run:
+    - run: &install_packages
         name: Install system packages
         command: |
           # python-snappy compression relies on C bindings to libsnappy-dev.
           # pyspark needs to run Java, so we install openjdk.
           apt-get update
           apt-get install -y libsnappy-dev openjdk-8-jre-headless
-    - run:
-        name: Run tests
+    - run: &run_tox_environment_matching_circleci_job_name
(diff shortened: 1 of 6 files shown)


jklukas:
Changed several things and squashed into one commit, so asking for review again:

- Removed several readthedocs references
- Removed the Makefile logic and refactored CircleCI config to use the most popular executable I could find for handling the ugly git mangling necessary to push a directory to a branch (using an npm module in this case)
- Added an ssh key fingerprint and tested that the docs-deploy can be successfully run from CircleCI

wlach:
Can you update the rtd site to redirect to ghpages once you've landed this?

PR #240, opened 2018-11-05
gh-pages fixes

jklukas:
Just a few more changes to fix observed problems with docs and CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,7 @@ jobs:
             git config user.email "fx-data-platform@mozilla.com"
             git config user.name "CircleCI docs-deploy job"
             npm install -g --silent gh-pages@2.0.1
-            gh-pages -d docs/_build/html
+            gh-pages --dotfiles --message "[skip ci] Updates" --dist docs/_build/html
 
   lint:
     docker:
--- a/tox.ini
+++ b/tox.ini
@@ -20,12 +20,9 @@ skipsdist = True
 usedevelop = True
 extras = testing
 commands = pytest {posargs:tests/}
-
-# 2018-09-14: moto isn't working with latest boto3, so we need to pin version for tests;
-# see https://github.com/spulec/moto/pull/1801
-# TODO: Remove this once moto restores compatibility
-deps =
-    boto3==1.7.84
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -111,7 +111,7 @@ jobs:
             git config user.email "fx-data-platform@mozilla.com"
             git config user.name "CircleCI docs-deploy job"
             npm install -g --silent gh-pages@2.0.1
-            gh-pages -d docs/_build/html
+            gh-pages --dotfiles --message "[skip ci] Updates" --dist docs/_build/html
```
jklukas: The `--dotfiles` flag is necessary for the `.nojekyll` file to get committed. The `[skip ci]` message prevents CircleCI from trying to build the gh-pages branch after this commit (which we expect to fail because it doesn't have the source code).

tox.ini
```diff
@@ -21,12 +21,6 @@ usedevelop = True
 extras = testing
 commands = pytest {posargs:tests/}
 
-# 2018-09-14: moto isn't working with latest boto3, so we need to pin version for tests;
-# see https://github.com/spulec/moto/pull/1801
-# TODO: Remove this once moto restores compatibility
-deps =
-    boto3==1.7.84
-
```
jklukas: The moto issue is now fixed, so we can remove this restriction.

jklukas:
Spoke too soon on the moto fix. I'm investigating now.

tox.ini
```diff
@@ -20,12 +20,9 @@ skipsdist = True
...
 commands = pytest {posargs:tests/}
-
-# 2018-09-14: moto isn't working with latest boto3, so we need to pin version for tests;
-# see https://github.com/spulec/moto/pull/1801
-# TODO: Remove this once moto restores compatibility
-deps =
-    boto3==1.7.84
+setenv =
+    AWS_SECRET_ACCESS_KEY=dummy_secret_needed_for_moto_tests
+    AWS_ACCESS_KEY_ID=dummy_key_needed_for_moto_tests
```
jklukas: Follows new example in https://github.com/spulec/moto/pull/1907

PR #242, opened 2018-11-21
Use threads instead of processes in Dataset.summaries

tdsmith:
Dataset.summaries uses a concurrent.futures.ProcessPoolExecutor to fetch multiple files from S3 at once.
ProcessPoolExecutor uses multiprocessing underneath, which defaults to using fork() on Unix.
Using fork() is dangerous and prone to deadlocks: https://codewithoutrules.com/2018/09/04/python-multiprocessing/

This is a possible source of observed deadlocks during calls to Dataset.records.

Using threads should not be a performance regression since the operation we're parallelizing over is network-bound,
not CPU-bound, so there should not be much contention for the GIL.

### DIFF
--- a/moztelemetry/dataset.py
+++ b/moztelemetry/dataset.py
@@ -137,7 +137,7 @@ class Dataset:
         datasets
         :param prefix: a prefix to the
         :param clauses: mapping of fields -> callables to refine the dataset
-        :param max_concurrency: number of processes to spawn when collecting S3 summaries,
+        :param max_concurrency: number of threads to spawn when collecting S3 summaries,
         defaults to 1.5 * cpu_count
         """
         self.bucket = bucket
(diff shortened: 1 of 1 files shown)


tdsmith:
I did a run with this branch and it didn't fail: https://dbc-caf9527b-e073.cloud.databricks.com/#job/753/run/3

That's not strong evidence that I actually fixed anything, since this job is flaky but sometimes succeeds.

I can keep my job pinned on this branch for a week or so if you'd like to shake it out, though I don't think it should break anything.

jklukas:
> I can keep my job pinned on this branch for a week or so if you'd like to shake it out, though I don't think it should break anything.

If it's not a significant inconvenience, let's do that. Then we avoid the confusion of rolling back the change if we find it's not sufficient to make performance predictable.

tdsmith:
There have been no additional related failures since pinning this branch on the 21st: https://dbc-caf9527b-e073.cloud.databricks.com/#job/715

(the failure of run 47 is because I was in the middle of revising the notebook when the job triggered, oops)

tdsmith:
🎉  Thanks!

jklukas:
You should be able to update the library in databricks once it's published.

jklukas:
Now published: https://pypi.org/project/python_moztelemetry/0.10.5/

